PR #12, opened 2025-05-06
ref and eval expansion support

theoephraim:
supporting expansion similar to dotenv-expand. This allows users to reference other items, and execute shell commands. For example:
- `FOO=$BAR`
- `FOO=${BAR}`
- `FOO=$(whoami)`
- `MULTIPLE=hello-$OTHER-$(whoami)`

Under the hood, we’ve refactored how the parser and functions work, so that function arguments can be nested function calls. The string-based expansion is then interpreted and translated into function calls, which can then be resolved along with any other resolvers.

For example:
`pre-${OTHER}-$(whoami)` is actually just expanded into `concat(“pre-“, ref(OTHER), “-“, eval(“whoami”))`

This makes the system extremely flexible, and in cases where escaping gets confusing, the user can always just use the function calls directly instead.


packages/env-spec-parser/src/classes.ts
```diff
@@ -291,6 +297,14 @@ export class ParsedEnvSpecConfigItem {
     return regularComments.map((comment) => comment.contents).join('\n');
   }
 
+  processExpansion(opts?: {}) {
+    if (this.data.value instanceof ParsedEnvSpecStaticValue) {
```
theoephraim: currently we are only expanding when the value is a single string (e.g. `ITEM="foo-$OTHER"`)

We may want to just expand everything, meaning something like `ITEM=concat("foo-", "$OTHER")` would also be expanded properly.

packages/env-spec-parser/src/expand.ts
```diff
@@ -0,0 +1,133 @@
...
+      name: 'concat',
+      args: new ParsedEnvSpecFunctionArgs({
+        values: partsWithExpandedVars,
+      }),
+    });
+  }
+}
+
+
+export function expandStaticValue(staticVal: ParsedEnvSpecStaticValue, opts?: {}) {
```
theoephraim: this could use some work to clean up, but its working

PR #18, opened 2025-05-13
curl installer

theoephraim:
adds a bash install script, meant to be consumed via curl, via some redirecting/shortened url, for example:
`curl -sSfL https://varlock.dev/install.sh | sh -s`

The script is fetching our [homebrew tap/formula](https://github.com/dmno-dev/homebrew-tap/blob/main/Formula/varlock.rb) to determine the current latest version. A simpler method used by other projects is  to rely on github's "latest" release, but because we release multiple packages from our monorepo, and changesets creates a release for each, there is no guarantee that the "latest" will be a varlock release. Fetching from our homebrew formula is a little weird, but it will keep the 2 install methods in sync, and we are already updating that file with the latest version number. Fetching it from anywhere else would mean keeping yet another thing in sync.

One other option would be fetching that version number from npm, but the npm package gets published via changesets _before_ the binary release workflow runs - so there is a short time where these 2 things may not be in sync.

Some other notes
- if homebrew is detected, it will install via homebrew, unless `--force-no-brew` flag is set
- default install dir is `~/.varlock/bin` (we anticipate shoving other stuff in there)

future improvements to the install script could be
- check if current version is already installed
- confirm if user wants to install globally, especially if we detect a package.json file in the current directory
- cleanup if install fails
- more error checks and messages

### DIFF
--- a/packages/varlock/install.sh
+++ b/packages/varlock/install.sh
@@ -8,50 +8,56 @@ set -e
 GITHUB_URL="https://github.com/dmno-dev/varlock"
 GITHUB_RELEASES_URL="${GITHUB_URL}/releases"
 HOMEBREW_FORMULA_URL="https://raw.githubusercontent.com/dmno-dev/homebrew-tap/refs/heads/main/Formula/varlock.rb"
+HOMEBREW_TAP_NAME="dmno-dev/tap/varlock"
 
 OS=""
 ARCH=""
 VERSION=""
 LATEST_VERSION=""
+INSTALL_DIR="${HOME}/.varlock/bin"
+INSTALL_DIR_UNEXPANDED="~/.varlock/bin"
+REINSTALL=""
+FORCE_NO_BREW="false"
 
 usage() {
-    require_cmd cat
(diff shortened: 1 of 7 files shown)


philmillman (approved):
lgtm 🚀 
we may want to an action that does a quick smoke test to install it and maybe run a command to verify the installation

packages/varlock/install.sh
```diff
@@ -0,0 +1,215 @@
...
+    -b, --bindir <DIR_PATH>     Sets bindir or installation directory. Defaults to ./bin
+
+ARGS:
+    <version>       is a specific varlock version from ${GITHUB_RELEASES_URL}. (defaults to latest)
+EOF
+    exit 2
+}
+
+parse_args() {
+  BINDIR=${BINDIR:-./bin}
```
philmillman: I think this should be an absolute path, otherwise it's just going to create a /bin wherever the install script is run

philmillman:
should we default to add it to their PATH as well?

PR #47, opened 2025-07-07
Add @defaultRequired=infer and associated tests

### LINKED ISSUE #44: new `@defaultRequired` infer option
### Description

Allow inferring whether an item is required based on if a value has been set for the item in `.env.schema`

### Motivation

Often in a schema you will set a static value for a non-sensitive item but forget to mark it as required. The act of setting the static value implicitly shows the item will be populated. In the rare case a default exists, but it might be cleared by another environment, you can of course still mark the item as `@optional`.

This is especially noticeable because the auto-generated types mark optional items as being possibly undefined.

### Proposed Solution

allow `# @defaultRequired=infer` to enable this mode.

### Alternatives

- manually mark items
- always enable this behaviour without explicitly opting it, could be only if no `@defaultRequired` is set

### Additional Information

Questions: 
- Is required inferred to be true only if a _static_ value is set `ITEM=foo`? Or if _any_ value is set `ITEM=someFn(...)`?
- does this only happen if a value is set in `.env.schema`, or does a value in `.env` also trigger this?

philmillman:
closes #44

### DIFF
--- a/packages/env-graph/src/lib/config-item.ts
+++ b/packages/env-graph/src/lib/config-item.ts
@@ -12,6 +12,7 @@ import {
 
 import { EnvGraphDataSource } from './data-source';
 import { ResolvedValue, Resolver } from './resolver';
+import { StaticValueResolver } from './resolver';
 
 export type ConfigItemDef = {
   description?: string;
@@ -130,15 +131,27 @@ export class ConfigItem {
   get isRequired() {
     for (const def of this.defs) {
       const defDecorators = def.itemDef.decorators || {};
-      if ('required' in defDecorators) {
-        return defDecorators.required.simplifiedValue;
-      } else if ('optional' in defDecorators) {
-        return !defDecorators.optional.simplifiedValue;
-      } else if ('defaultRequired' in def.source.decorators) {
-        return def.source.decorators.defaultRequired.simplifiedValue;
+
+      // Explicit per-item decorators
+      if ('required' in defDecorators) return true;
+      if ('optional' in defDecorators) return false;
+
+      // Root-level @defaultRequired
+      if ('defaultRequired' in def.source.decorators) {
+        const val = def.source.decorators.defaultRequired.simplifiedValue;
+        if (val === 'infer') {
+          const resolver = def.itemDef.resolver;
+          if (resolver instanceof StaticValueResolver) {
+            return resolver.staticValue !== undefined && resolver.staticValue !== '';
+          } else {
+            return true; // function value
+          }
+        }
+        return val; // explicit true or false
       }
     }
-    return true; // otherwise default to true
+    // defaults to true
+    return true;
   }
 
   get isSensitive() {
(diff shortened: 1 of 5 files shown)


packages/env-graph/src/lib/config-item.ts
```diff
@@ -130,15 +131,27 @@ export class ConfigItem {
...
       const defDecorators = def.itemDef.decorators || {};
-      if ('required' in defDecorators) {
-        return defDecorators.required.simplifiedValue;
-      } else if ('optional' in defDecorators) {
-        return !defDecorators.optional.simplifiedValue;
-      } else if ('defaultRequired' in def.source.decorators) {
-        return def.source.decorators.defaultRequired.simplifiedValue;
+
+      // Explicit per-item decorators
+      if ('required' in defDecorators) return true;
```
theoephraim: This isnt quite right because you can do `@required=false` or `@optional=false`

packages/env-graph/src/lib/config-item.ts
```diff
@@ -270,13 +283,12 @@ export class ConfigItem {
         const validationError = new ValidationError(`Unexpected non-error thrown during validation - ${err}`);
         validationError.cause = err;
         this.validationErrors = [validationError];
-      }
+      }
```
theoephraim: turn on lint on save. if youre really against it, we can try to set up lint-staged

packages/env-graph/test/root-decorators.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  {
+    label: 'function value is required',
+    headers: '# @defaultRequired=infer',
+    values: 'BAR=fnCall()',
+    expected: { BAR: true },
+  },
+  {
+    label: 'explicit optional overrides infer',
+    headers: '# @defaultRequired=infer',
+    values: '# @optional\nBAZ=qux\n\nBAR=val',
```
theoephraim: might be worth installing [outdent](https://www.npmjs.com/package/outdent) and then breaking these into multiple lines?

packages/env-graph/test/root-decorators.test.ts
```diff
@@ -0,0 +1,91 @@
+import { describe, it, expect } from 'vitest';
```
theoephraim, line 1: could just make this the _required_ test suite rather than root decorators?

packages/env-graph/test/root-decorators.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        for (const key in expected) {
+          const item = g.configSchema[key];
+          expect(item.isRequired).toBe(expected[key]);
+        }
+      });
+    });
+  };
+}
+
+describe('@defaultRequired root decorator', requiredInferenceTests([
```
theoephraim, line 31: probably want a test for no default set

packages/env-graph/test/root-decorators.test.ts
philmillman: yeah this format could definitely use some love

packages/env-graph/src/lib/config-item.ts
philmillman: it should already be on by default from the `.vscode` setting in the repo. I think this might have been a symptom of how I reverted some changes, will investigate

packages/env-graph/test/root-decorators.test.ts
philmillman: maybe let's reorg once we have a few more tests

PR #48, opened 2025-07-07
refactor loading setup

theoephraim:
many changes to the core as part of building out the nextjs integration.

- importing typed env is now always via `import { ENV } from 'varlock/env'`
- clean up data source  labels on env files  (these will be displayed by nextjs later)
- automatically patch globals to enable security features, rather than asking user to call functions. This is required to avoid hoisting and initialization order problems.
- add new root level decorators to let user disable log redaction and leak prevention. These are exposed in the serialized graph
- add type generation for `process.env`! (same intellisense but everything is a string)
- add/clean up log redaction and leak prevention runtime code


philmillman (approved):
LGTM - i'm assuming the patching is similar to DMNo so I didn't go over that with a fine toothed comb

PR #51, opened 2025-07-07
fix preview release action

theoephraim:
our github action for pkg.pr.new releases was is not working for our versioned release PRs. Changesets is not able to detect which packages to release in the PRs where we are removing all the changeset files.

Instead we detect that this is a versioned release branch, and then look for updated package.json files.


philmillman (approved):
🙏

PR #52, opened 2025-07-08
@defaultSensitive=inferFromPrefix(...) option and associated tests

### LINKED ISSUE #43: new `@defaultSensitive` inferFromPrefix option
### Description

Allow inferring if an item is sensitive based on whether its key begins with a specific prefix. This would allow users coming from frameworks where prefixes are used to mark items as "public" to continue using that prefix to control item sensitivity.

For example, a Next.js project could use `# @defaultSensitive=inferFromPrefix(NEXT_PUBLIC_)`

### Motivation

We want the incremental adoption path for users to be as simple as possible. While our _sensitive_ settings are more flexible, as a first step, some users may want to just continue relying on the prefixes they are used to.

### Proposed Solution

Modfiy the root level `@defaultSensitive` decorator to allow a value of `inferFromPrefix(PREFIX)` which is a function call containing a single arg which is the prefix to use.

### Alternatives

You can explicitly mark every item that differs from the default of true/false.

One other option to consider is whether we automatically detect the framework and use its known behaviour.

### Additional Information

_No response_

philmillman:
closes #43

### DIFF
--- a/packages/env-graph/src/lib/config-item.ts
+++ b/packages/env-graph/src/lib/config-item.ts
@@ -170,7 +170,19 @@ export class ConfigItem {
         return defDecorators.sensitive.simplifiedValue;
         // TODO: do we want an opposite decorator similar to @required/@optional -- maybe @public?
       } else if ('defaultSensitive' in def.source.decorators) {
-        return def.source.decorators.defaultSensitive.simplifiedValue;
+        const dec = def.source.decorators.defaultSensitive;
+        // Handle function call: inferFromPrefix(PREFIX)
+        if (dec.value instanceof ParsedEnvSpecFunctionCall && dec.value.name === 'inferFromPrefix') {
+          const args = dec.value.simplifiedArgs;
+          // Accepts a single string prefix as first arg
+          const prefix = Array.isArray(args) && args.length > 0 ? args[0] : undefined;
+          if (typeof prefix === 'string' && this.key.startsWith(prefix)) {
+            return false; // Not sensitive if matches prefix
+          }
+          return true; // Sensitive otherwise
+        }
+        // Fallback to static/boolean value
+        return dec.simplifiedValue;
       }
     }
     return true;
--- a/packages/env-graph/test/sensitive-decorator.test.ts
+++ b/packages/env-graph/test/sensitive-decorator.test.ts
@@ -32,22 +32,33 @@ function sensitiveInferenceTests(
 
 describe('@defaultSensitive inferFromPrefix', sensitiveInferenceTests([
   {
-    label: 'key matches prefix is not sensitive',
+    label: 'base case @defaultSensitive=inferFromPrefix',
     headers: outdent`# @defaultSensitive=inferFromPrefix(PUBLIC_)`,
     values: outdent`
       PUBLIC_FOO=bar
+      BAR=baz
+    `,
+    expected: { PUBLIC_FOO: false, BAR: true },
+  },
+  {
+    label: 'key matches prefix is not sensitive (with explicit override)',
+    headers: outdent`# @defaultSensitive=inferFromPrefix(PUBLIC_)`,
+    values: outdent`
+      PUBLIC_FOO=bar
+      # @sensitive=true
       SECRET_BAR=baz
     `,
     expected: { PUBLIC_FOO: false, SECRET_BAR: true },
   },
   {
-    label: 'key does not match prefix is sensitive',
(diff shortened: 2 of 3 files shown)


packages/env-graph/test/sensitive-decorator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    label: 'key matches prefix is not sensitive',
+    headers: outdent`# @defaultSensitive=inferFromPrefix(PUBLIC_)`,
+    values: outdent`
+      PUBLIC_FOO=bar
+      SECRET_BAR=baz
+    `,
+    expected: { PUBLIC_FOO: false, SECRET_BAR: true },
+  },
+  {
+    label: 'key does not match prefix is sensitive',
```
theoephraim: I was initially confused about the difference between these first 2 tests. Kind of covers both if you have one matching and one not?

packages/env-graph/test/sensitive-decorator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    headers: outdent`# @defaultSensitive=inferFromPrefix(PUBLIC_)`,
+    values: outdent`
+      # @sensitive=false
+      SECRET_BAR=baz
+      # @sensitive=true
+      PUBLIC_FOO=bar
+    `,
+    expected: { SECRET_BAR: false, PUBLIC_FOO: true },
+  },
+  {
```
theoephraim, line 74: maybe add another to each of these 2 to check overriding the default

packages/env-graph/src/lib/config-item.ts
```diff
@@ -170,7 +170,19 @@ export class ConfigItem {
...
         // TODO: do we want an opposite decorator similar to @required/@optional -- maybe @public?
       } else if ('defaultSensitive' in def.source.decorators) {
-        return def.source.decorators.defaultSensitive.simplifiedValue;
+        const dec = def.source.decorators.defaultSensitive;
+        // Handle function call: inferFromPrefix(PREFIX)
+        if (dec.value instanceof ParsedEnvSpecFunctionCall && dec.value.name === 'inferFromPrefix') {
+          const args = dec.value.simplifiedArgs;
+          // Accepts a single string prefix as first arg
+          const prefix = Array.isArray(args) && args.length > 0 ? args[0] : undefined;
+          if (typeof prefix === 'string' && this.key.startsWith(prefix)) {
```
theoephraim, line 179: Def skip for now, but eventually we probably want to trigger a schema error (at the root level, not on an item) if you pass bad args into inferFromPrefix. Ideally we take care of that with something more general, where we register available decorators with the available kinds of values/args.

packages/env-graph/test/sensitive-decorator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+      });
+    });
+  };
+}
+
+describe('@defaultSensitive inferFromPrefix', sensitiveInferenceTests([
+  {
+    label: 'key matches prefix is not sensitive',
+    headers: outdent`# @defaultSensitive=inferFromPrefix(PUBLIC_)`,
```
theoephraim, line 36: this is totally a nit and up to you, but I wonder if it might be cleaner to just have the headers and items together in a single prop.  explicitly writing the divider versus  and extra key and call to outdent :shrug: 
```
  envFile: outdent`
    # @defaultSensitive=inferFromPrefix(PUBLIC_)
    # ---
    PUBLIC_FOO=bar
    SECRET_BAR=baz
 `
 ```

File: packages/env-graph/test/sensitive-decorator.test.ts [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+  {
+    label: 'static @defaultSensitive=false still works',
+    headers: outdent`# @defaultSensitive=false`,
+    values: outdent`
+      FOO=bar
+      BAR=baz
+    `,
+    expected: { FOO: false, BAR: false },
+  },
+]));
```
theoephraim, line 101: probably good to have one that verify the default when no `@defaultSensitive` decorator is present

theoephraim (changes requested):
this is great :D Thanks for implementing it!

packages/env-graph/test/sensitive-decorator.test.ts
philmillman: My thinking here was that eventually we'll have helpers to populate this or generate permutations of each type of options for the various tests. It seemed to make sense to keep these separate for these reasons, but I understand in its current form that a single block is more readable

PR #54, opened 2025-07-08
add docs for new root decorators and options

philmillman:


packages/varlock-website/src/content/docs/reference/root-decorators.mdx
```diff
@@ -25,26 +25,66 @@ APP_ENV=development
...
 <div>
-### `@defaultRequired  `
-Sets the the default behavior of each item being _required_. Can be overridden on individual items using [`@required`](/reference/item-decorators/#required) or [`@optional`](/reference/item-decorators/#optional).
+### `@defaultRequired`
+Sets the default behavior of each item being _required_. Can be overridden on individual items using [`@required`](/reference/item-decorators/#required) or [`@optional`](/reference/item-decorators/#optional).
+
+**Options:**
+- `true` (default): All items are required unless marked optional.
+- `false`: All items are optional unless marked required.
+- `infer`: Items with a static or function value are required; items with an empty string or no value are optional. Can be overridden per item.
```
theoephraim:
```suggestion
- `infer`: Items with a value set in `.env.schema` will be required; items with an empty string or no value are optional. Can be overridden per item.
```

theoephraim: I realized we'll probably want to make sure that it only infers based on the schema. Otherwise the required-ness could change based on which env you are loading.

packages/varlock-website/src/content/docs/reference/root-decorators.mdx
```diff
@@ -25,26 +25,66 @@ APP_ENV=development
...
+# @required
+REQUIRED_ITEM= # required (explicit)
 ```
 </div>
 
 <div>
-### `@defaultSensitive  `
-Sets the the default state of each item being treated sensitive. Can be overridden on individual items using [`@sensitive`](/reference/item-decorators/#sensitive).
+### `@defaultSensitive`
+Sets the default state of each item being treated as sensitive. Can be overridden on individual items using [`@sensitive`](/reference/item-decorators/#sensitive).
```
theoephraim:
```suggestion
Sets the default state of each item being treated as _sensitive_. Can be overridden on individual items using [`@sensitive`](/reference/item-decorators/#sensitive).
```

packages/varlock-website/src/content/docs/reference/root-decorators.mdx
```diff
@@ -25,26 +25,66 @@ APP_ENV=development
...
+
+**Options:**
+- `true` (default): All items are sensitive unless marked otherwise.
+- `false`: All items are not sensitive unless marked otherwise.
+- `inferFromPrefix(PREFIX)`: Items whose key starts with the given `PREFIX` are _not_ sensitive; all others are sensitive. Useful for marking e.g. `PUBLIC_` keys as non-sensitive by default.
 
 ```env-spec
-# @defaultSensitive=false
+# @defaultSensitive=inferFromPrefix(PUBLIC_)
```
theoephraim, line 63: should we call this `inferFromPublicPrefix`? Or since its always a public prefix do you think its ok?

packages/varlock-website/src/content/docs/reference/root-decorators.mdx
```diff
@@ -25,26 +25,66 @@ APP_ENV=development
...
-SERVICE_X_AUTH_TOKEN=
+
+PUBLIC_FOO=bar   # not sensitive
+SECRET_BAR=baz   # sensitive
+
+# @sensitive=false
+FOO=bar          # not sensitive (explicit)
+
+# @sensitive=true
+PUBLIC_BAR=baz   # sensitive (explicit)
```
theoephraim:
```suggestion
PUBLIC_FOO=  # not sensitive (due to matching prefix)
OTHER_FOO=  # sensitive (default when prefix does not match)

# @sensitive=true
PUBLIC_BAR=  # sensitive (explicit decorator overrides prefix)
# @sensitive=false
OTHER_BAR=  # not sensitive (explicit)
```

packages/varlock-website/src/content/docs/reference/root-decorators.mdx
```diff
@@ -25,26 +25,66 @@ APP_ENV=development
...
+```
+</div>
+
+<div>
+### `@disable`
+Skips loading config items from a file or data source. If present (or set to `true`), the file is ignored and no items are loaded from it. Useful for conditionally disabling a schema or env file.
```
theoephraim:
```suggestion
Skips loading config items from a file or data source. If true, the file is ignored and no items are loaded from it. Useful for conditionally disabling a schema or env file.
```

theoephraim: saying "if present" sounds like it also includes `@disable=false`

packages/varlock-website/src/content/docs/reference/root-decorators.mdx
```diff
@@ -25,26 +25,66 @@ APP_ENV=development
...
+
+```env-spec
+# @disable
```
theoephraim:
```suggestion
# @disable  # (shorthand for @disable=true)
```

packages/varlock-website/src/content/docs/reference/root-decorators.mdx
philmillman: it's tricky because it's kind of inferring it based on the absence of a prefix (if that makes sense)

philmillman: maybe let's leave it for now and we can revisit if there is confusion

philmillman:
> some suggestions, but up to you... should do a follow up that makes sure `@defaultRequired=infer` only applies to values set in `.env.schema`

https://github.com/dmno-dev/varlock/issues/55

PR #56, opened 2025-07-08
env-flag-handling

theoephraim:
update how we handle setting our env flag

- no longer check duruing load, instead during resolution (this avoids errors during varlock init)
- allow passing in an env flag via cli arg - necessary to match next.js default behaviour
- resolve env flag item early, before loading other env files
- introduce new `remap()` function so we can transform a branch name injected by an external PaaS / CI into a custom env flag


packages/env-graph/src/lib/config-item.ts
```diff
@@ -197,16 +203,23 @@ export class ConfigItem {
       return;
     }
 
-    if (this.isResolved) throw new Error('item already resolved');
+    if (this.isResolved) {
+      // throw new Error('item already resolved');
```
philmillman:
```suggestion

```

packages/env-graph/src/lib/config-item.ts
```diff
@@ -92,6 +92,12 @@ export class ConfigItem {
   }
 
   async process() {
+    // we add the final override def here so that if we process the item early (like when resolving our envFlag) we will always it
```
philmillman:
```suggestion
    // we add the final override def here so that if we process the item early (like when resolving our envFlag) we will always have it
```

philmillman (approved):
two super tiny nits, nice job! :shipit:

PR #58, opened 2025-07-09
use varlock in varlock-website code

theoephraim:
very basic setup of using varlock within the website.

we’ll still need to think through a deeper astro integration to get better DX, but at least we are dogfooding a little, which should help us test end-to-end that everything is working properly.


packages/env-graph/src/lib/env-graph.ts
```diff
@@ -166,6 +166,9 @@ export class EnvGraph {
         // process and resolve any other items our env flag depends on
         for (const depKey of envFlagItem.valueResolver?.deps || []) {
```
theoephraim, line 167: found an edge case in the special envFlag early resolution... Will likely want to clean that up a bit so its not quite as special, but this is ok for now.

philmillman (approved):
LGTM! 🐶

PR #65, opened 2025-07-11
next integration smoke tests

theoephraim:
Testing our integrations is tricky. We have to set up a real next repo and run real commands. There are also so many permutations of settings, node versions, package versions, and other things to test. Also many things need to be tested one at a time, because a bad key or a leak will fail a build.

This is just super simple start to this kind of testing. We'll likely want to extract some of these helpers as we'll want to have similar tests for each framework integration.

### DIFF
--- a/packages/integrations/nextjs/.gitignore
+++ b/packages/integrations/nextjs/.gitignore
@@ -1,0 +1,1 @@
+test/test-project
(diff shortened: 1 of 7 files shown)


theoephraim:
closing this now that we have some smoke tests. Might reference the branch again when expanding the next tests

PR #77, opened 2025-07-15
vite integration

theoephraim:
initial work on a varlock + vite integration - which is a [vite plugin](https://vite.dev/guide/api-plugin).

Some notes
- vite automatically does some loading of .env files, including loading env-specific files and handling expansion. It seems our plugin overrides this without any issues, but we'll want to write some tests around this case.
- using vite's `config.define` causes some issues because it will not replace explicitly imported things meaning `ENV.X` will not be replaced if we `import { ENV } from 'varlock/env'`). Instead we rely on a [rollup replace plugin](https://www.npmjs.com/package/@rollup/plugin-replace) under the hood

Some other refactoring was required along the way:
- the default vite example uses `tsc` which was checking some of the varlock code due to how it was being built
- enabled `"verbatimModuleSyntax": true` and adjusted our imports
- no longer re-exporting a type from `@env-spec/env-graph`, due to dts generation not being unable to inline the package into the final bundle the same way it does for the js files
- new options to enable nicer error messages when using the `ENV` proxy
- exposing the paths of .env file sources


packages/integrations/vite/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+  });
+}
+
+// we run this right away so the globals get injected into the vite.config file
+reloadConfig();
+
+export function varlockVitePlugin(
+  options?: {
+    /**
+     * option to bundle DMNO_CONFIG as well, which will include all config items
```
philmillman: copy/pasta?

packages/varlock-website/src/content/docs/integrations/vite.mdx
```diff
@@ -0,0 +1,169 @@
...
+<div class="page-badges">
+  <Badge npmPackage="@varlock/vite-integration" />
+</div>
+
+While many frameworks use Vite under the hood, some projects use Vite directly. When doing so, Vite does do some [automatic loading of .env files](https://vite.dev/guide/env-and-mode.html), but it is fairly limited.
+
+To integrate varlock into a Vite application, you must use our [`@varlock/vite-integration`](https://www.npmjs.com/package/@varlock/vite-integration) package, which is a [Vite plugin](https://vite.dev/guide/using-plugins.html).
+
+:::note[Vite-powered frameworks]
+This integration is intended for use with projects that are using Vite directly. For frameworks that use Vite under the hood, you may need a specific integration for it.
```
philmillman:
```suggestion
This integration is intended for use with projects that are using Vite directly. For frameworks that use Vite under the hood, you may need a framework-specific integration.
```

packages/varlock-website/src/content/docs/integrations/vite.mdx
```diff
@@ -0,0 +1,169 @@
...
+      plugins: [ varlockVitePlugin(), /* other plugins... */ ]
+    });
+    ```
+</Steps>
+
+---
+
+## Accessing environment variables
+
+You can continue to use `import.meta.env.SOMEVAR` as usual, but we recommend using Varlock's imported `ENV` object for better type-safety and improved developer experience:
```
philmillman:
```suggestion
You can continue to use `import.meta.env.SOMEVAR` as usual, but we recommend using varlock's imported `ENV` object for better type-safety and improved developer experience:
```

packages/varlock-website/src/content/docs/integrations/vite.mdx
```diff
@@ -0,0 +1,169 @@
...
+import { defineConfig } from 'vite';
+import { varlockVitePlugin } from '@varlock/vite-integration';
++import { ENV } from 'varlock/env';
+
++doSomethingWithEnvVar(ENV.FOO);
+
+export default defineConfig({ /* ... */ });
+```
+
+⚠️ If you find you are not getting type completion on `ENV`, you may need to add your vite config and generated type files (usually `env.d.ts`) to your `tsconfig.json`'s `include` array.
```
philmillman: should probably use a real callout here

philmillman (approved):
a few nits

:shipit:

PR #79, opened 2025-07-16
next integration + docs improvements

theoephraim:
some small tweaks and docs improvements based on user feedback


packages/varlock-website/src/content/docs/integrations/nextjs.mdx
```diff
@@ -69,8 +74,11 @@ Turbopack does not yet provide any plugin system, so using the config plugin is
     />
 
 1. **Override `@next/env` with our drop-in replacement**
+    
+    Next.js does not have apis we can hook into, so we must override their internal .env-loading package.
```
philmillman:
```suggestion
    Next.js does not have APIs we can hook into, so we must override their internal .env-loading package.
```

PR #84, opened 2025-07-23
improve anonymous telemetry setup

theoephraim:
Many improvements to our telemetry setup, based on inspiration from [astro](https://www.npmjs.com/package/@astrojs/telemetry)

- rename opt out env var
- automatically disable telemetry if running a local dev build
- track current varlock version number (and if it is an actual release vs preview build
- collect very basic system/os/node/platform info
- use random UUID to identify users, stored in `.varlock/config.json` rather than grouping all under id `anonymous` (better for posthog)
- collect anonymized project id
- graceful shutdown and exit hooks, to give telemetry request a little time to finish
- smoothly handle telemetry request failure (and being offline)


scripts/update-homebrew-formula.js
```diff
@@ -51,7 +51,7 @@ class Varlock < Formula
   end
 
   test do
-    system bin/"varlock", "--post-install", "brew"
+    assert_equal "${VERSION}", shell_output("#{bin}/varlock --post-install brew").strip
```
theoephraim, line 54: I still need to test this!

packages/varlock/src/cli/helpers/telemetry.ts
```diff
@@ -0,0 +1,267 @@
...
+  debug('gitRemoteUrl (will be anonymized)', gitRemoteUrl);
+  return anonymizeValue(gitRemoteUrl);
+}
+
+
+type TelemetryMeta = {
+  // project info
+  anonymous_project_id: string | null;
+  // version information
+  node_version: string;
```
philmillman, line 153: should we separate bundled node (standalone) vs "system" node (npm)?

philmillman (approved):
one small comment, looks great otherwise!

PR #91, opened 2025-07-26
fix vite reloading logic, small init bugfixes

theoephraim:
- fixes vite reloading logic
- fixes issue when running varlock init in a pnpm monorepo root
- fixes crash when git is not installed
- swaps out rollup plugin replace for ast-based strategy
- specific fixes for vue script+setup with vite plugin


packages/utils/src/git-utils.ts
```diff
@@ -5,9 +5,14 @@ export async function checkIsFileGitIgnored(path: string, warnIfNotGitRepo = fal
     await asyncExec(`git check-ignore ${path} -q`);
     return true;
   } catch (err) {
+    const stderr = (err as any).stderr as string;
+    // git is not installed, so we don't know
+    if (stderr.includes('command not found')) return undefined;
```
theoephraim, line 10: main difference here is that we are returning undefined to signify that we don't know. That difference between false and undefined is not used anywhere yet, but it feels a bit safer for the future.

PR #93, opened 2025-07-27
add case where there's no git cli

### LINKED ISSUE #90: [BUG]: `varlock init` fails if git CLI is not available
Ran `npx varlock init` on stackblitz, and got a crash due to `git` cli not being installed.
I believe it is doing some checks to determine which .env files may be git ignored.

It should not cause a crash.

philmillman:
closes #90


theoephraim:
Have this covered already in #91

PR #98, opened 2025-07-29
add internal exports for env-graph and associated

philmillman:


theoephraim (approved):
lgtm. you could possibly have this stuff exported as `varlock/internal` rather than in an `internal` const. Not really sure yet what will feel better

philmillman:
> lgtm. you could possibly have this stuff exported as `varlock/internal` rather than in an `internal` const. Not really sure yet what will feel better

I'll defer to you there.

PR #121, opened 2025-08-07
add root decorator docs for opt out of preventLeaks and redactLogs

### LINKED ISSUE #120: [docs] add @redactLogs and @preventLeaks
There are also opt-out root decorators, which need to be added to the docs.
`@redactLogs=false` and `@preventLeaks=false`

_Originally posted by @theoephraim in https://github.com/dmno-dev/varlock/discussions/118#discussioncomment-14036623_

philmillman:
closes #120


packages/varlock-website/src/content/docs/reference/root-decorators.mdx
```diff
@@ -121,4 +121,54 @@ Enables automatic type generation based on your schema.
...
+**Options:**
+- `true` (default): Leak detection is enabled
+- `false`: Leak detection is disabled (useful for debugging)
+
+```env-spec
+# @preventLeaks=false
+# ---
+
+# @sensitive
+POSTHOG_API_KEY=my-secret-value
```
theoephraim: maybe not the best example only because posthog project api keys are public, but not really a huge deal!

philmillman: I just did that because I'm lazy and it would match the screenshot, I'll update

PR #125, opened 2025-08-12
MVP varlock GitHub action

### LINKED ISSUE #95: [Feature Request] GitHub Action
### Description

An official GitHub Action would allow a more secure way to use .env files without hard-coded secrets.

### Motivation

_No response_

### Proposed Solution

_No response_

### Alternatives

_No response_

### Additional Information

_No response_

philmillman:
closes #95


packages/varlock-github-action/README.md
```diff
@@ -0,0 +1,236 @@
...
+      - name: Load environment variables
+        uses: dmno-dev/varlock-github-action@v1
+        with:
+          working-directory: './config'
+          environment: 'production'
+          show-summary: 'true'
+          fail-on-error: 'true'
+          output-format: 'env'
+      
+      - name: Use loaded variables
```
theoephraim: prob dont need this step, and if you do have it, include an env var in the example

packages/varlock-github-action/README.md
```diff
@@ -0,0 +1,236 @@
...
+        run: |
+          echo "Environment loaded successfully"
+```
+
+## Inputs
+
+| Input | Description | Required | Default |
+|-------|-------------|----------|---------|
+| `working-directory` | Directory containing @env-spec files | No | `.` |
+| `environment` | Environment to load (e.g., production, development) | No | - |
```
theoephraim: this is a bit confusing. The user should be passing in their env flag as an env var using the name of their env flag.
We could potentially still preserve this option to mirror what we have on the CLI, but it should be clear that it's not used in most cases.

packages/varlock-github-action/README.md
```diff
@@ -0,0 +1,236 @@
...
+| `show-summary` | Show a summary of loaded environment variables | No | `true` |
+| `fail-on-error` | Fail the action if validation errors are found | No | `true` |
+| `output-format` | Format for environment variable output (env, json) | No | `env` |
```
theoephraim: how is the output used?

packages/varlock-github-action/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Outputs
+
+| Output | Description |
+|--------|-------------|
+| `summary` | Summary of loaded environment variables |
+| `error-count` | Number of validation errors found |
+| `warning-count` | Number of validation warnings found |
+
+## @env-spec Environment File Detection
```
theoephraim: what does this mean? How is ordering significant?

packages/varlock-github-action/README.md
philmillman: that can be reworked a bit similar to the dmno action where we either output individual env vars or one big json blob, it's more relevant for multi-stage workflows where the env vars don't persist

packages/varlock-github-action/README.md
philmillman: agreed

packages/varlock-github-action/action.yml
```diff
@@ -0,0 +1,39 @@
...
+    value: ${{ steps.varlock.outputs.summary }}
+  error-count:
+    description: 'Number of validation errors found'
+    value: ${{ steps.varlock.outputs.error-count }}
+  json-env:
+    description: 'JSON blob containing all environment variables (only available when output-format is "json"). This will contain all environment variables, including sensitive ones. Be careful.'
+    value: ${{ steps.varlock.outputs.json-env }}
+
+runs:
+  using: 'node20'
```
theoephraim: node 24?

packages/varlock-github-action/src/index.ts
```diff
@@ -0,0 +1,265 @@
...
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+export function checkForEnvFiles(workingDir: string): boolean {
+  try {
+    const files = readdirSync(workingDir);
+    const envFiles = files.filter((file: string) => file.startsWith('.env'));
```
theoephraim: probably want to check for `.env` and `.env.*`

packages/varlock-github-action/src/index.ts
```diff
@@ -0,0 +1,265 @@
...
+        stdio: 'pipe',
+        encoding: 'utf8',
+      }).toString();
+
+      return { output, errorCount: 0 };
+    } catch (error: any) {
+      if (error.stdout) {
+        const output = error.stdout.toString();
+        // Parse error count from output if available
+        const errorCount = (output.match(/error/gi) || []).length;
```
theoephraim: this seems a bit sketchy. Probably want to check some more precise json format?

packages/varlock-github-action/test/sensitive-data.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        NODE_ENV: { value: 'development', isSensitive: false },
+      },
+    };
+
+    // Test the function - it should not throw any errors
+    setEnvironmentVariables(envGraph);
+
+    // The function doesn't return anything, but we can verify it was called
+    // by checking that no errors were thrown
+    expect(true).toBe(true);
```
theoephraim: do you need any expect here? If it throws above, we'll get an error.

packages/varlock-github-action/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "lib": ["ES2023"],
+    "strict": true,
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "verbatimModuleSyntax": true,
+    "customConditions": ["ts-src"],
+    "outDir": "./dist",
+    "rootDir": "./src"
+  },
+  "include": ["src/**/*", "test/index.integration.test.ts", "test/index.test.ts"],
```
theoephraim: `test/**/*`?

theoephraim (approved):
overall looks good. Im still a bit confused about the output format stuff, and the tests could definitely be better, but should be ok to start.

packages/varlock-github-action/action.yml
philmillman: probably best to match varlock which is 22?

Issue #119, opened 2025-08-07
[Feature Request] conditionally required items


theoephraim:
### Description

Currently we can only mark items as required or not - in all cases. We'd like to be able to mark things as required under certain conditions.

The most common case would be based on an env flag ("item x is required in production only") but one can imagine making it totally flexible.

### Proposed Solution

Add new capabilities to the `@required` decorator. 

Long-term we may want a fully flexible syntax with many functions (ex: `@required=not(eq($OTHER_VAR, 'asdf'))`), but this opens up some larger questions, since our functions currently are only used for item values, not for decorator values. Also only a subset of our functions would make sense in this context.

Considering 90% of cases will be based on the current environment (dev/prod), I think we can get away with something that checks the env flag only. We also know this will be resolved early, so it makes things simpler, and it further formalizes the envFlag as important.
 
- it should probably use the value function syntax, similar to `@type=string(startsWith=abc)` -- `@required=env(production)`
- could be called `env` `forEnv` `ifEnv` `isEnv` `envFlag`
- it should accept multiple envs -- `@required=env(preview, production)`

theoephraim:
env-based dynamic required implemented by #136 

eventually we may implement more general logic

PR #136, opened 2025-08-27
@required - env() helper

theoephraim:
First steps towards https://github.com/dmno-dev/varlock/issues/119

This adds a new helper to set item required-ness based on the current env (as set by @envFlag)

Some notes:
- not sure about the `env()` name, could be `ifEnv` `isEnv` `forEnv`
- value can be a single env, or multiple
- works for both `@required` and `@optional`
- if the value is dynamic, we must respect that in type generation

In an ideal world, we'd handle any arbitrary logic here, and likely unify our resolver functions (which are used for item values) with our decorator values. However this introduces a lot of complexity, because now items can be dependent on each other in new ways, and the resolution process would need to handle that.

For now, we are using a much more limited subset. Using the envFlag is totally safe, because we always resolve that up front, so it does not introduce any new ordering / dependency issues. We likely want to add a few more helpers here, but they will likely have additional limitations, which will allow us to keep things a bit simpler.

### DIFF
--- a/.changeset/thin-parents-lick.md
+++ b/.changeset/thin-parents-lick.md
@@ -2,4 +2,4 @@
 "varlock": patch
 ---
 
-add new `env()` helper for @required decorator, to allow dynamically setting required-ness based on current env
+add new `forEnv()` helper for @required decorator, to allow dynamically setting required-ness based on current env flag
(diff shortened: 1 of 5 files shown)


philmillman:
I think I like `forEnv(...)` as it seems the most semantic

philmillman (approved):
🚢

Issue #140, opened 2025-08-29
[Feature Request] bun plugin


philmillman:
### Description

Because `bun` relies on `NODE_ENV` to load certain env files (see: https://bun.sh/guides/runtime/set-env) it has unintended effects when you use `envFlag` in certain situations (see: #139) 

### Motivation

We need `envFlag` to work predictably no matter the runtime environment. 

### Proposed Solution

Create a `bun` plugin https://bun.sh/docs/bundler/plugins the ensures `NODE_ENV` is not overriding any intended behaviour. 

### Alternatives

custom logic + manual `NODE_ENV` overrides

trevordcampbell:
I was noodling around a bit in the Varlock Codebase and figured I would throw up a PR #141 for an alternative approach of baking the fix into Varlock core. Feel free to discard it if you want to stick with the Plugin route, but this modification does seem to be working with bun.

You can do a quick test of the modified Varlock CLI executable in this ready-to-rip container:
https://github.com/trevordcampbell/varlock-testing

philmillman:
Thanks @trevordcampbell, will review with @theoephraim and see what makes the most sense!

theoephraim:
see https://varlock.dev/integrations/bun/

Issue #139, opened 2025-08-28
[BUG]: Bun auto-loading .env files conflicting with varlock loaded env


trevordcampbell:
### varlock version

0.0.11

### Steps to reproduce

1. **Clone repo:** https://github.com/trevordcampbell/varlock-testing

2. Open project in **DevContainer** _(Already fully configured)_

3. **Install:** `bun install`

4. **Run the script** with all possible environments:

```zsh
# Test 1: Load command with production env
APP_ENV=production bun varlock load
# Expected:
# ✅ APP_ENV*: "production"
# ✅ LOG_LEVEL*: "warn"
# ✅ OVERRIDE_LOADED_FROM*: "production"
# Actual output:
# ✅ APP_ENV*: "production"
# ✅ LOG_LEVEL: "debug"
# ✅ OVERRIDE_LOADED_FROM*: "development"

# Test 2: Run command with production env
APP_ENV=production bunx varlock run -- bun script.ts
# Expected output:
# APP_ENV: production
# LOG_LEVEL: warn
# OVERRIDE_LOADED_FROM: production
# Actual output:
# APP_ENV: production
# LOG_LEVEL: debug
# OVERRIDE_LOADED_FROM: development

# Test 3: Run command with staging env
APP_ENV=staging bunx varlock run -- bun script.ts
# Expected output:
# APP_ENV: staging
# LOG_LEVEL: info
# OVERRIDE_LOADED_FROM: staging
# Actual output:
# APP_ENV: staging
# LOG_LEVEL: debug
# OVERRIDE_LOADED_FROM: development
```

### What is expected?

When setting `APP_ENV=production`, Varlock should:
1. Load `.env.production` override file
2. Apply production-specific environment variables
3. Show `OVERRIDE_LOADED_FROM=production`

### What is actually happening?

Regardless of `APP_ENV` value, Varlock:
1. Always loads `.env.development` override file
2. Applies development-specific environment variables
3. Always shows `OVERRIDE_LOADED_FROM=development`

### System Info

```shell
System:
    OS: Linux 6.15 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (14) arm64 unknown
    Memory: 12.12 GB / 15.66 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  npmPackages:
    @biomejs/biome: ^2.2.2 => 2.2.2 
    @types/bun: latest => 1.2.21 
[...]

PR #141, opened 2025-08-30
PR: Make Varlock authoritative under Bun by neutralizing Bun dotenv and hardening resolver

trevordcampbell:
# Summary
- Fixes: Bun autoloading of dotenv can override Varlock’s resolved values, causing “development” to leak into non-dev runs.
- Approach: Make Varlock the single source of truth for env resolution and neutralize Bun’s dotenv via `--env-file`, while preserving runtime-agnostic semantics and keeping UX simple.
- Result: Works for any `@envFlag` value/name, is zero-config for most users, and remains portable to other runtimes.

References:
- Bug report: [#139](https://github.com/dmno-dev/varlock/issues/139)
- Original proposal for a Bun plugin: [#140](https://github.com/dmno-dev/varlock/issues/140)
- Bun env handling docs: [Bun environment variables](https://bun.com/docs/runtime/env)

---

## Why this bug happens under Bun
Bun auto-loads dotenv files based on `NODE_ENV` (in precedence order): `.env` → `.env.(development|production|test)` → `.env.local`. It does not recognize custom names (e.g., `.env.staging`). See [Bun environment variables](https://bun.com/docs/runtime/env).

Varlock’s model resolves from `.env.schema` + `.env.<@envFlag>` (e.g., `APP_ENV`) regardless of `NODE_ENV`. When Bun autoloads before or alongside Varlock, values from `.env.development` and/or `.env.local` can override the intended values, especially when `NODE_ENV` is defaulting to development. This is the core of [#139](https://github.com/dmno-dev/varlock/issues/139).

---

## Why a Varlock-core fix was proposed instead of a Bun plugin
- Deterministic and supported: Bun’s `--env-file` deterministically controls dotenv behavior; passing an empty file results in no autoload. This is stable and documented in [Bun environment variables](https://bun.com/docs/runtime/env).
- Zero-config UX: Users keep using `varlock load` / `varlock run`. No per-project preloads or bunfig changes.
- Ordering guarantees: Injecting Varlock’s resolved env into the spawned process happens at process creation time—so early module init sees the intended values.
- Runtime-agnostic and maintainable: Works with Bun today, and the same pattern can adapt to other runtimes that autoload env files. Less coupling to Bun internals than a plugin. See [#140](https://github.com/dmno-dev/varlock/issues/140).

---

## What changed (code and behavior)

- Harden resolver to ignore ambient `process.env` for schema-defined keys by default
  - File: `packages/varlock/env-graph/lib/config-item.ts`
  - Gate `process.env` overrides via `EnvGraph.respectExistingEnv`:
    - If an item exists in the schema, ambient `process.env` will not override it unless `respectExistingEnv` is true.
    - Special case: allow the `@envFlag` key (e.g., `APP_ENV`) to be set from `process.env` during the early pass so Varlock can select the correct `.env.<env>` files from the start.
  - Effect: Prevents Bun’s preloaded values from clobbering schema-defined keys, without blocking setting the environment flag via `process.env`.

- Default handling for `.env.local` is opt-out
  - File: `packages/varlock/env-graph/lib/loader.ts`
  - We include `.env.local` and `.env.<env>.local` by default (consistent with existing Varlock behavior).
  - Added `excludeLocal?: boolean` to disable locals when explicitly requested.
  - Propagation to CLI:
    - `packages/varlock/src/lib/load-graph.ts` plumbs `excludeLocal` and `respectExistingEnv`.
    - `packages/varlock/src/cli/commands/load.command.ts` adds flags:
      - `--exclude-local`: Exclude `.env.local` and `.env.[env].local`
      - `--respect-existing-env`: Allow ambient `process.env` to override schema-defined keys
    - `packages/varlock/src/cli/commands/run.command.ts` mirrors the same flags for the run path.

- Bun-aware runner
  - File: `packages/varlock/src/cli/commands/run.command.ts`
  - Detects `bun`/`bunx` and neutralizes Bun dotenv by passing an empty `--env-file`.
  - Injects Varlock’s resolved environment into the child process using a hardened pass-through:
    - Default: whitelist essential OS vars (`PATH`, `HOME`, `SHELL`, `TERM`, `TZ`, `LANG`, `LC_ALL`, `PWD`, `TMPDIR`, `TEMP`, `TMP`) + all Varlock-resolved keys.
    - This minimizes ambient leakage and avoids accidental overrides by pre-existing parent env keys.
  - Optional `--bun-sync-node-env`: sets `NODE_ENV` to the resolved `@envFlag` value when enabled. Off by default to avoid surprises.
  - Notes:
    - Resolver strictness and `.env.local` defaults apply to all commands (Node or Bun).
    - Bun dotenv neutralization only applies when the child command is `bun` or `bunx`.

---

## Backward compatibility and UX

- `import { ENV } from 'varlock/env'` still works
  - Runtime autoload flows (`varlock/auto-load`) are unchanged. `ENV` is still initialized from `process.env.__VARLOCK_ENV`, injected by `varlock run` or integrations, and exposes resolved values to application code.

- Default behavior changes you should be aware of
  - Ambient overrides: By default, ambient `process.env` values no longer override schema-defined keys. Restore old behavior with `--respect-existing-env` (or a future `.varlockrc` default).
  - `.env.local`: Included by default, consistent with the current Varlock mental model. You can explicitly opt-out via `--exclude-local`.
  - Bun-only: When the child command is Bun/Bunx, `varlock run` disables Bun dotenv to ensure no drift from Varlock’s resolved env.

- Performance
  - No material changes in the core resolution performance. The whitelist-based env pass-through is trivial overhead.

---

## How it works (step-by-step)

1) Resolution
- Varlock loads `.env.*` files in the project directory, parses `.env.schema`, discovers `@envFlag` (e.g., `APP_ENV`).
- During an early pass, the env flag key is resolved. If `APP_ENV` is defined in the ambient environment, it is respected for this key only to select the appropriate `.env.<env>` files.
- Schema-defined keys are then resolved from `.env.schema`, the environment-specific file(s), and optionally `.env.local` per flag. Ambient `process.env` is ignored for these keys by default (unless `--respect-existing-env` is set).

2) Running with Bun
- If `varlock run -- bun ...` or `varlock run -- bunx ...`:
  - Varlock spawns Bun with `--env-file <empty-temp-file>`, neutralizing Bun’s dotenv mechanism.
  - Varlock injects the resolved env and a small whitelist of parent env vars into the child process, ensuring libraries see the intended values at module init.
  - Optionally syncs `NODE_ENV` to the resolved `@envFlag` value if `--bun-sync-node-env` is provided.

---

## New Flags and defaults
- `--exclude-local`: Exclude `.env.local` and `.env.[env].local`. Default: not excluded (included).
- `--respect-existing-env`: Allow ambient `process.env` to override schema-defined keys. Default: false.
- `--bun-sync-node-env`: When running Bun, set `NODE_ENV` to the resolved `@envFlag` value. Default: false.

---

## Code references (selected)
- Resolver hardening and env flag special-case
  - `packages/varlock/env-graph/lib/config-item.ts`:
    - Applies `process.env` overrides only when `respectExistingEnv` is true or the item is not in the schema.
    - Always allows `process.env` to set the env flag key (e.g., `APP_ENV`) during the early pass so `.env.<env>` selection works.

- .env file loading, opt-out `.env.local`
  - `packages/varlock/env-graph/lib/loader.ts`:
    - Discovers `.env.*` files.
    - Includes locals by default; disables them only if `excludeLocal === true`.

- CLI flags
  - `packages/varlock/src/cli/commands/load.command.ts`: `--exclude-local`, `--respect-existing-env`, `--env`.
  - `packages/varlock/src/cli/commands/run.command.ts`: `--exclude-local`, `--respect-existing-env`, `--bun-sync-node-env`, `--env`.

- Bun-aware runner
  - `packages/varlock/src/cli/commands/run.command.ts`:
    - Detects `bun`/`bunx`.
    - Prepends `--env-file <empty-temp-file>`.
    - Constructs a whitelist pass-through env for the child process.
    - Optional `NODE_ENV` sync from `@envFlag`.

---

## Testing instructions

A ready-to-run Bun-based DevContainer test project (with updated executables) is available for reviewers:

### [varlock-testing DevContainer](https://github.com/trevordcampbell/varlock-testing)

It contains:
- A set of `.env*` files (`.env.schema`, `.env.production`, `.env.development`, `.env.local`, etc.)
- A simple `script.ts` that logs selected variables using `process.env` (do not import `varlock/env` for this specific test path).

Uses an SEA bundle of the modified Varlock CLI (no install required in the test project):
- From project root, run:

Basic resolution (JSON):
```bash
bun ./varlock-cli-executable-bun-fix.cjs load --format json
```

Run a Bun script with dotenv neutralized:
```bash
bun ./varlock-cli-executable-bun-fix.cjs run -- bun script.ts
```

Select staging env (env flag via ambient):
```bash
APP_ENV=staging bun ./varlock-cli-executable-bun-fix.cjs -- bun script.ts
```

Opt-out `.env.local`:
```bash
APP_ENV=production bun ./varlock-cli-executable-bun-fix.cjs run --exclude-local -- bun script.ts
```

Allow ambient process.env to override schema-defined keys (off by default):
```bash
OVERRIDE=from-ambient APP_ENV=production bun ./varlock-cli-executable-bun-fix.cjs load --respect-existing-env --format json | jq .OVERRIDE
```

Sync NODE_ENV to your `@envFlag` value (optional):
```bash
APP_ENV=production bun ./varlock-cli-executable-bun-fix.cjs run --bun-sync-node-env -- bun script.ts
```

Expected behavior:
- With `APP_ENV=production`, values come from `.env.schema` + `.env.production` (+ `.env.local` unless excluded).
- With `APP_ENV=staging`, values come from `.env.schema` + `.env.staging` (+ `.env.local` unless excluded).
- `OVERRIDE` only comes from ambient when `--respect-existing-env` is used.
- Bun dotenv does not alter results when running through `varlock run`.

---

## Maintainability and portability
- Runtime-agnostic: The pattern (resolve first, neutralize external dotenv, inject resolved env) can be applied to other CLIs/runtimes that auto-load dotenv.
- Minimal surface area: Lives inside Varlock’s CLI and resolver core. No per-project plugin/preload to maintain.
- Safe-by-default: Whitelist pass-through for parent env reduces leakage; explicit flags control deviations.
- Clear escape hatches: `--respect-existing-env`, `--exclude-local`, `--bun-sync-node-env`.

---

## Compatibility notes and migration guidance
- Ambient overrides for schema-defined keys now default to “off.”
  - If you rely on ambient overrides, add `--respect-existing-env` to your CLI invocations (or adopt a config default when available).
- `.env.local` remains included by default for local dev parity. Use `--exclude-local` if you want to disable it for specific runs.

---

## Acknowledgements and prior art
- This PR addresses [#139](https://github.com/dmno-dev/varlock/issues/139), which documents the Bun autoload override problem.
- It considers and chooses a core Varlock fix over a Bun plugin as initially suggested in [#140](https://github.com/dmno-dev/varlock/issues/140) for the reasons outlined above.
- References:
  - Bun env docs: [Bun environment variables](https://bun.com/docs/runtime/env)

---

## Proposed .varlockrc (optional, future docs)

```json
{
  "resolver": {
    "respectExistingEnv": false
  },
  "bun": {
    "dotenv": "none",
    "syncNodeEnv": false
  }
}
```

This example captures the defaults introduced in this PR and shows how a future `.varlockrc` could be used to centralize project-wide preferences. We can add this to the docs as a forward-looking example.


theoephraim:
Closing now that bun has better options to disable its env loading

PR #157, opened 2025-09-26
Leak detection fastify tweak

theoephraim:
fastify does some funky handling of content-type so the header is not set yet when we check it during leak detection. So we adjusted the logic to still run the leak scan if the body is a string and no content type is present


philmillman (approved):
🚢

PR #158, opened 2025-09-26
allow setting envFlag from imported file

theoephraim:
We now allow setting envFlag via an imported file.

This is useful if you had many services and wanted to import that common setting from a single place.

To make this happen, when we encounter an `@envFlag` decorator, we now set the envFlag on the data source itself, and walk up the parent chain setting it if it is empty. We also (still) walk up the parent chain when checking for an envFlag.


philmillman (approved):
:shipit:

PR #162, opened 2025-10-01
fix import path issue

theoephraim:
fixes path issue related to imports. It was not causing problems in most situations, but when there are chains of multiple imports, it can create a problem. Hard to test this one, because in our test suite, we mock the file loading using virtual files and follow a different code path. In the future we could think about mocking fs methods, but its likely not worth the effort.


philmillman (approved):
🚢

PR #163, opened 2025-10-01
fix varlock not found when running node directly

theoephraim:
Received reports of `varlock not found` - meaning our executable was not found by our library code which was trying to invoke it to load env vars.

This was due to `execSync('varlock ...')` not being able to find varlock in the PATH. This was surprising because this works in most cases. Tracked it down to the difference between running via a package manager vs directly. The package managers automatically inject `./node_modules/.bin` into the PATH.

This PR adjusts things so that if varlock is not found, we walk up the ancestor directories looking for node_modules, and when found, we try to use the full path.

Along the way, also improved handling of all code that handles not finding an executable, and all code that walks up directories to the root.

### DIFF
--- a/.changeset/tangy-numbers-nail.md
+++ b/.changeset/tangy-numbers-nail.md
@@ -1,0 +1,7 @@
+---
+"@varlock/nextjs-integration": patch
+"@varlock/vite-integration": patch
+"varlock": patch
+---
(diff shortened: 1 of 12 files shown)


philmillman (approved):
:shipit:

PR #168, opened 2025-10-14
plugins system + 1pass plugin

theoephraim:
A ton going on here. Ideally would have broken this up more, but it was in figuring out the DX for plugin authoring that a lot of this became clear, so authoring the plugin, while building the plugin system, and the general changes around decorators, all together was how it was actually built. 

---

- creates a new plugin system, which can load plugins using a new `@plugin` decorator. These can be loaded from local files, installed node modules, https, or from npm
- plugins will be able to register new root decorators, item decorators, data types, and resolver functions
- plugins will be installed globally in the graph, no notion of limiting scope
- formalizes decorators, and unifies their arguments with the existing value resolver system
  - this is because many plugins will now be referencing existing config items, so we want it to be the same as referencing items within values
  - this also enables arbitrary logic for required/sensitive/etc -- for example `@required=eq($SOME_VAR, 'foo')`
  - have added a few utility resolvers (if, eq) and more to come
- first version of a 1password plugin, along with docs


example-monorepo/packages/plugin-test/.env.op
```diff
@@ -0,0 +1,32 @@
+# for now we'll import directly via local paths in node_modules
+# but soon we'll add ways of importing using package name and likely relying on package.json
+# # @plugin(./node_modules/@varlock/simple-vault-plugin/dist/plugin.js)
+# # @plugin(node_modules:@varlock/simple-vault-plugin)
+# # @plugin(npm:@varlock/simple-vault-plugin@1.2.3)
+#
+# # @initSimpleVault(id=vault1, key=SIMPLE_VAULT_ENCRYPTION_KEY)
+#
+# @plugin(./node_modules/@varlock/1password-plugin/dist/plugin.js)
```
theoephraim: we'll have an easier to way to reference installed node modules rather than using a path like this

example-monorepo/packages/plugin-test/.env.op
```diff
@@ -0,0 +1,32 @@
...
+# # @plugin(https://varlock-plugins-registry.dmno.workers.dev/varlock-1password-plugin-0.0.0.tgz)
+# @initOpVault(id=v1, token=$OP_TOKEN, enableCliAuth=true)
```
theoephraim: most plugins will probably introduce a root decorator that will be responsible for initialization - which usually means setting options, and wiring up auth to env items. In many cases you may want to create multiple instances of a plugin like this.

Some plugins may not need one? Or could optionally rely on defaults if no init decorator was used

example-monorepo/packages/plugin-test/.env.op
```diff
@@ -0,0 +1,32 @@
...
+# # @sensitive
+# ENCRYPTED_ITEM=simpleVault(vault1, GAdBm3dAnsR2+KUBaBMwaQhN+tM8Tq8Bk/nEk74CHAJE0E+aUDAIDYtwc5k8lec=)
+
+
+# @sensitive
+OP_TOKEN=
+
+OP_VAULT_NAME=example
+
+OP1=op(v1, "op://dev test/${OP_VAULT_NAME}/credential")
```
theoephraim: Example of new custom resolver. It is referencing the vault ids in the `initOpVault` calls. Nothing special or formalized involved though... plugins can track their own state and do whatever they want.

They can use their module singleton scope to do so. In the future, we'll likely add more hooks, so they could for example initialize or reset their state as part of the overall loading process

In this 1pass example, the vault IDs are optional, with id on both the init call and items defaulting to `_default` if not specified.

Note that we'll get schema errors before value resolution if ids are no good, because we require the ids to be static strings, so are evaluated early.

example-monorepo/packages/plugin-test/.env.schema
```diff
@@ -0,0 +1,27 @@
...
+# ---
+
+# our environment flag
+# @type=enum(dev, preview, prod, test)
+# @docsUrl=https://example.com/docs/app_env
+APP_ENV=dev
+
+# some description
+# @required=true
+# @docs(https://example1.com)
```
theoephraim, line 15: `@docsUrl` is now deprecated in favor of a `@docs()` function, which can be called multiple times. It also now supports both a URL and optional label. Multiple docs links across all definitions will all be combined together and included in generated types.

example-monorepo/packages/plugin-test/.env.schema
```diff
@@ -0,0 +1,27 @@
+# @defaultRequired=infer @defaultSensitive=false
+# @generateTypes(lang=ts, path=env.d.ts)
+#
+# @currentEnv=$APP_ENV
```
theoephraim, line 3: `@envFlag=APP_ENV` is now deprecated (although still works) in favor of `@currentEnv=$APP_ENV`

While this now uses the unified resolvers to reference another item, we add some limitations - it must be set to `ref(SOME_VAR)` (or the unexpanded equivalent), and SOME_VAR must be an item that is defined in the same file.

Doesn't change the DX in this case very much, but I do think as we have more of these that unifying the resolvers will make the overall DX much better.

packages/env-spec-parser/src/classes.ts
```diff
@@ -299,26 +300,32 @@ export class ParsedEnvSpecBlankLine {
...
   constructor(public data: {
     key: string;
     value: ParsedEnvSpecStaticValue | ParsedEnvSpecFunctionCall | undefined;
     preComments: Array<ParsedEnvSpecDecoratorComment | ParsedEnvSpecComment>;
     postComment: ParsedEnvSpecDecoratorComment | ParsedEnvSpecComment | undefined;
     _location?: any;
-  }) {}
+  }) {
+    if (this.data.value) {
+      const expanded = expand(this.data.value!);
```
theoephraim, line 313: expansion now happens automatically within the parser library, rather than needing to be called explicitly

packages/plugins/1password/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    "./plugin": "./dist/plugin.cjs"
+  },
+  "files": ["dist"],
+  "scripts": {
+    "dev": "pnpm run copy-wasm && tsup --watch",
+    "build": "tsup && pnpm run copy-wasm",
+    "copy-wasm": "mkdir -p dist && cp node_modules/@1password/sdk-core/nodejs/core_bg.wasm ./dist",
```
theoephraim, line 20: 1pass internal lib uses wasm and we need to copy it into our final dist folder, since we dont have any dependencies being loaded.

This is what nudged me in the direction of plugins usually being a folder with a package.json, rather than defaulting to a single file. We may still want to support the single file case for when a user wants to just do a little custom stuff, but any published plugins should be a full folder.

packages/plugins/1password/src/plugin.ts
```diff
@@ -0,0 +1,218 @@
...
+import { createDeferredPromise, DeferredPromise } from '@env-spec/utils/defer';
+import { Client, createClient } from '@1password/sdk';
+import { opCliRead } from './cli-helper';
+
+const { ValidationError, SchemaError, ResolutionError } = plugin.ERRORS;
+
+const PLUGIN_VERSION = plugin.version;
+const OP_ICON = 'simple-icons:1password';
+
+plugin.name = '1pass';
```
theoephraim, line 12: plugin is injected globally... its a bit awkward, but the best way I've found so far to inject in memory objects into dynamically imported code. Otherwise we'd end up re-bundling copies of our classes and functions into the plugin, and instanceof checks would not work.

packages/plugins/simple-vault/package.json
```diff
@@ -0,0 +1,55 @@
+{
```
theoephraim: ignore this whole thing for now. Need to figure out how we will deal with CLIs and plugins before this plugin can be useful.

example-monorepo/packages/plugin-test/.env.op
theoephraim: it could be something like `@plugin(node_modules:@varlock/1password-plugin)`
or it could just be the default when no other protocol is provided `@plugin(@varlock/1password-plugin)`

- local paths will start with `./` or `../` or `/`
- registries (npm/jsr/?) will be `npm:packageName@1.2.3`

example-monorepo/packages/plugin-test/.env.op
philmillman: I think maybe they should all have it for consistency's sake, and the case where there's some sort of update which adds parameters to a previously "empty" init

example-monorepo/packages/plugin-test/.env.schema
philmillman: is the new name to give us time to deprecate `envFlag`?

example-monorepo/packages/plugin-test/.env.schema
philmillman: makes sense to me. I think since there are multiple deprecations in this release we should probably consider a proper minor (0.1.0?) release

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+    } else if (errMessage.includes("isn't a vault in this account")) {
+      // message looks like -- "asdf" isn't a vault in this account...
+      // so we will extract the vault name/id
+      const matches = errMessage.match(/"([^"]+)" isn't a vault in this account/);
+      const vaultNameOrId = matches?.[1] || 'unknown';
+      return new ResolutionError(`1password vault "${vaultNameOrId}" not found in account connected to op cli`, {
+        code: 'BAD_VAULT_REFERENCE',
+        extraMetadata: { badVaultId: vaultNameOrId },
+        tip: [
+          'By not using a service account token, you are relying on your local 1password cli installation and authentication.',
```
philmillman:
```suggestion
          'By not using a service account token, you are relying on your local 1Password CLI installation and authentication.',
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+      const matches = errMessage.match(/could not find item (.+) in vault (.+)/);
+      const itemNameOrId = matches?.[1] || 'unknown';
+      const vaultId = matches?.[2] || 'unknown';
+
+      // const vaultNameOrId = errMessage.substring(1, errMessage.substring(1).indexOf('"') + 1);
+      return new ResolutionError(`1password item "${itemNameOrId}" not found in vault "${vaultId}"`, {
+        code: 'BAD_ITEM_REFERENCE',
+        extraMetadata: { badItemId: itemNameOrId, vaultId },
+        tip: [
+          'Double check the item in your 1password vault.',
```
philmillman:
```suggestion
          'Double check the item in your 1Password vault.',
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+      });
+    } else if (errMessage.includes(' does not have a field ')) {
+      // message includes `item 'dev test/example' does not have a field 'bad field name'`
+      const matches = errMessage.match(/item '([^']+)' does not have a field '([^']+)'/);
+      const itemNameOrId = matches?.[1] || 'unknown';
+      const [vaultId, itemId] = itemNameOrId.split('/');
+      const fieldNameOrId = matches?.[2]?.replace('.', '/') || 'unknown';
+
+      // const vaultNameOrId = errMessage.substring(1, errMessage.substring(1).indexOf('"') + 1);
+      return new ResolutionError(`1password vault "${vaultId}" item "${itemId}" does not have field "${fieldNameOrId}"`, {
```
philmillman:
```suggestion
      return new ResolutionError(`1Password vault "${vaultId}" item "${itemId}" does not have field "${fieldNameOrId}"`, {
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+
+
+
+    // when the desktop app integration is not connected, some interactive CLI help is displayed
+    // however if it dismissed, we get an error with no message
+    // TODO: figure out the right workflow here?
+    if (!errMessage) {
+      return new ResolutionError('1password cli not configured', {
+        tip: [
+          'By not using a service account token, you are relying on your local 1password cli installation and authentication.',
```
philmillman:
```suggestion
          'By not using a service account token, you are relying on your local 1Password CLI installation and authentication.',
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+          'You many need to enable the 1password Desktop app integration, see https://developer.1password.com/docs/cli/get-started/#step-2-turn-on-the-1password-desktop-app-integration',
```
philmillman:
```suggestion
          'You many need to enable the 1Password Desktop app integration, see https://developer.1password.com/docs/cli/get-started/#step-2-turn-on-the-1password-desktop-app-integration',
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+          'Try running `op whoami` to make sure the cli is connected to the correct account',
```
philmillman:
```suggestion
          'Try running `op whoami` to make sure the CLI is connected to the correct account',
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+          'You may need to call `op signout` and `op signin` to select the correct account.',
+        ],
+      });
+    }
+    return new Error(`1password cli error - ${errMessage || 'unknown'}`);
+  } else if ((err as any).code === 'ENOENT') {
+    return new ResolutionError('1password cli `op` not found', {
+      tip: [
+        'By not using a service account token, you are relying on your local 1password cli installation for ambient auth.',
```
philmillman:
```suggestion
        'By not using a service account token, you are relying on your local 1Password CLI installation for ambient auth.',
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+        'But your local 1password cli (`op`) was not found. Install it here - https://developer.1password.com/docs/cli/get-started/',
```
philmillman:
```suggestion
        'But your local 1Password CLI (`op`) was not found. Install it here - https://developer.1password.com/docs/cli/get-started/',
```

packages/plugins/1password/src/cli-helper.ts
```diff
@@ -0,0 +1,321 @@
...
+      ],
+    });
+  } else {
+    return new ResolutionError(`Problem invoking 1password cli: ${(err as any).message}`);
```
philmillman:
```suggestion
    return new ResolutionError(`Problem invoking 1Password CLI: ${(err as any).message}`);
```

packages/plugins/1password/src/plugin.ts
```diff
@@ -0,0 +1,218 @@
...
+        return deferred.promise;
+      }
+    } else {
+      throw new SchemaError('1Password vault not properly initialized - must provide token or enableCliAuth');
+    }
+  }
+
+  private async executeReadBatch() {
+    const opClient = await this.opClientPromise;
+    if (!opClient) throw new Error('Expected op sdk to be initialized');
```
philmillman, line 88:
```suggestion
    if (!opClient) throw new Error('Expected op SDK to be initialized');
```

packages/plugins/1password/src/plugin.ts
```diff
@@ -0,0 +1,218 @@
...
+          } else {
+            dp.reject(new ResolutionError('bulk fetch is missing item response'));
+          }
+        }
+      }
+    } catch (err) {
+      let commonErr;
+      // 1pass sdk throws strings as errors...
+      if (typeof err === 'string') {
+        commonErr = new ResolutionError(`1password SDK error - ${err}`);
```
philmillman:
```suggestion
        commonErr = new ResolutionError(`1Password SDK error - ${err}`);
```

packages/plugins/1password/src/plugin.ts
```diff
@@ -0,0 +1,218 @@
...
+});
+
+
+plugin.registerDataType({
+  name: 'op/serviceAccountToken',
+  sensitive: true,
+  typeDescription: 'Service account token used to authenticate with the [1Password CLI](https://developer.1password.com/docs/cli/get-started/) and [SDKs](https://developer.1password.com/docs/sdks/)',
+  icon: OP_ICON,
+  docs: [
+    '1password service accounts',
```
philmillman:
```suggestion
    '1Password service accounts',
```

packages/plugins/1password/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@varlock/1password-plugin",
+  "description": "Varlock plugin to pull data from 1password",
```
philmillman:
```suggestion
  "description": "Varlock plugin to pull data from 1Password",
```

packages/plugins/simple-vault/package.json
philmillman: maybe cherry pick it into another branch/PR?

example-monorepo/packages/plugin-test/.env.schema
theoephraim: We could leave the name the same and still make it work, but I thought the decorator naming didnt really make as much sense. Like with `@envFlag` youre saying - the name of the env flag item is "APP_ENV". Whereas "$APP_ENV" really represents the value that it resolves to, so you're saying the `@currentEnv` is what `$APP_ENV` will resolve to (eg "dev")

example-monorepo/packages/plugin-test/.env.schema
theoephraim: So far my thinking has been that we bump up to 0.1.0 (or maybe 1.0.0?) when we reach a point where we are ready to guarantee stability, and otherwise just keep rolling with it. But open to other ideas. Should also probably write something about our versioning strategy somewhere in the docs?

example-monorepo/packages/plugin-test/.env.schema
philmillman: yeah makes sense to me, probably good for us to draft some guidelines about these types of deprecations though

example-monorepo/packages/plugin-test/.env.schema
philmillman: this feels like at minimum a minor version to me. we're definitely not adhering to semver currently but normally you'd announce the deprecation in a minor version and remove it in a major

example-monorepo/packages/plugin-test/.env.op
theoephraim: I think forcing the user to have an init call could make sense in the case where we standardize the init call (something like `@initPlugin`). Otherwise it's up to the plugin to do whatever it wants, which could be no init decorator, or several decorators involved in setting things up correctly.

If we do that, the question is whether you allow the plugin to introduce any behavior at all before calling init.

PR #172, opened 2025-10-19
WIP: `replace()` function

rennokki:
My use case is weird:

```env
BASE_URL="http://localhost:3000"

# I want to be able to do one of these:
SECURE_BASE_URL=replace(ref(BASE_URL), "http://", "https://")
SECURE_BASE_URL=replace(ref(BASE_URL), regex("^http://(.+)$"), "https://$1")

# http://localhost:3000 👉🏼 https://localhost:3000
```

I would need some advice if possible, the project is a bit tricky to navigate or it is because the day is ending and I am a bit tired 😅 I'm planning on adding more utility functions if I can get to sort this one.

### DIFF
--- a/.changeset/flat-turkeys-win.md
+++ b/.changeset/flat-turkeys-win.md
@@ -1,0 +1,6 @@
+---
+"@env-spec/parser": patch
+"varlock": patch
+---
+
+added `replace()` function to replace a string with support for ref()


packages/env-spec-parser/test/functions.test.ts
```diff
@@ -135,6 +146,51 @@ describe('ref expansion', functionValueTests({
   },
 }));
 
+describe('replace expansion', functionValueTests({
```
theoephraim: No need for this - as long as the ref stuff is tested and replace is tested, we can assume they will work correctly together.

packages/env-spec-parser/test/functions.test.ts
```diff
@@ -58,10 +58,21 @@ describe('function calls', functionValueTests({
...
   },
+  'replace()': {
+    input: 'ITEM=replace("foo", "f", "b")',
+    expected: { ITEM: 'boo' },
+  },
   'nested function calls - array': {
     input: 'OTHERVAL=d\nITEM=concat("a", fallback("", "b"), exec("echo c"), ref(OTHERVAL))',
     expected: { ITEM: 'abcd' },
   },
+  'nested function calls - replace()': {
```
theoephraim: dont need this

packages/varlock/env-graph/lib/resolver.ts
```diff
@@ -251,6 +251,42 @@ export class RefResolver extends Resolver {
...
+
+  private originalString?: string;
+  private searchString?: string;
+  private replacementString?: string;
+
+  async _process() {
+    if (this.fnArgs.length !== 1) {
+      throw new SchemaError('replace() expects three child args');
+    }
+    if (!(this.fnArgs[0] instanceof StaticValueResolver)) {
```
theoephraim, line 267: This is checking that the args are all static, as in `replace("foo", "f", "z")` but we want to support the args not being static - meaning they are other function calls (which includes using expansion) - ie `replace($SOMEVAR, 'find', $OTHER)`.

So within _process, we are basically just checking if args look right, but without necessarily knowing what the actual values are. (We only can check values if the arg was static). Then within _resolve, we can get the actual values, do some additional error checks, and then return our result.

packages/env-spec-parser/src/simple-resolver.ts
```diff
@@ -28,6 +28,16 @@ export function simpleResolver(
         } else {
           throw new Error('Invalid `ref` args');
         }
+      } else if (valOrFn.name === 'replace') {
```
theoephraim, line 31: We don't actually need this defined here at all. This is just a very simple implementation of a few functions which are used to test the parser itself.

packages/env-spec-parser/test/functions.test.ts
```diff
@@ -58,10 +58,21 @@ describe('function calls', functionValueTests({
       ITEM: 'foo-val',
     },
   },
+  'replace()': {
```
theoephraim, line 61: This file is testing how the parser itself handles parsing functions, and we don't need to define or test the new resolver within the parser. The tests we want to add are in [env-graph](https://github.com/dmno-dev/varlock/blob/main/packages/varlock/env-graph/test/resolvers.test.ts)

packages/varlock/env-graph/lib/resolver.ts
theoephraim: also note - this isnt getting tested yet because you had added the tests in the parser lib.

theoephraim:
@rennokki - just wanted to follow up here. You still interested in helping out with some of these utility functions? We'd love to have you on board :) and happy to help get things kicked off by pairing or provide any support we can. It should all be much clearer now that the new stuff merged.

theoephraim:
@rennokki - going to close this for now. Would still love your help adding more functions if you are interested :) There are more examples now and it should be more straightforward!

PR #181, opened 2025-10-29
Enable sitemap generation in robots.txt

philmillman:
required so the sitemap is discoverable in cloud AI search


philmillman:
> lgtm - did we just have it turned off initially before we wanted things indexed?

yeah I think we just forgot to turn back on

PR #200, opened 2025-11-12
add project-level config file to allow telemetry opt-out

theoephraim:
previously we checked the user's home folder for a `.varlock/config.json` file. This file can contain a `"disableTelemetry": true` to let the user opt out of telemetry.

We now also look for a project-level config file - by walking up the directory tree. These 2 config files will be merged, with project settings overriding user settings.

This allows a telemetry opt-out to live within the repo, and persist across team members.


philmillman (approved):
:shipit:

PR #224, opened 2025-12-04
CF workers overview docs

philmillman:


packages/varlock-website/src/content/docs/integrations/cloudflare.mdx
```diff
@@ -0,0 +1,219 @@
...
+    +    otherPlugin()
+    +  ],
+    });
+    ```
+
+    The varlock plugin will automatically detect Cloudflare Workers and use the `resolved-env` mode, which injects the fully resolved environment data into your built code.
+
+</Steps>
+
+### Local Development
```
theoephraim: this stuff is only needed for the non vite way!

From further reading, I guess there are actually 3 ways.

- use vite (and use our `ENV` object)
- inject the whole env and use our `ENV` object
- use the `"env -i` trick and dont use our `ENV` object

packages/varlock-website/src/content/docs/integrations/cloudflare.mdx
```diff
@@ -0,0 +1,219 @@
...
+
+:::note[Future improvement]
+Cloudflare is working on the ability to push secrets with a deploy command. See [this pull request](https://github.com/cloudflare/workers-sdk/pull/10896) for updates.
+:::
+
+---
+
+## Approach 2: Using Wrangler Vars and Secrets (Without Vite)
+
+If you prefer not to use the Vite plugin, you can use Wrangler's built-in environment variable handling. Note that this approach will not provide varlock's security features like leak detection and log redaction.
```
theoephraim: We actually can  provide the security stuff! It just isnt injected by default. I think we'll need to make a new import like `import 'varlock/init-only'` that does the rest of what auto-load does but without calling out to the cli.

packages/varlock-website/src/content/docs/integrations/cloudflare.mdx
```diff
@@ -0,0 +1,219 @@
...
+Varlock can load multiple _environment-specific_ `.env` files (e.g., `.env.development`, `.env.preview`, `.env.production`) by using the [`@currentEnv` root decorator](/reference/root-decorators/#currentenv).
+
+You can use Cloudflare's branch name to determine the environment:
+
+```env-spec title=".env.schema"
+# @currentEnv=$APP_ENV
+# ---
+WORKERS_CI_BRANCH=
+# @type=enum(development, preview, production, test)
+APP_ENV=remap($WORKERS_CI_BRANCH, production="main", preview=regex(.*), development=undefined)
```
theoephraim, line 165: I think I want to deprecate remap, or change it anyway... I really dont like the restrictions on keynames.

PR #228, opened 2025-12-09
add charset to content-type header

philmillman:
UTF8 chars in lllms-*.txt weren't being rendered properly

see discussion here https://github.com/delucis/starlight-llms-txt/issues/44


philmillman:
confirmed on preview: 
image

PR #237, opened 2025-12-27
feat: support Google Secret Manager Plugin

ya7010:
Since an SDK is not currently available, this is a PR for Google Secret Manager Plugin support to enable my team to adopt Varlock.

### DIFF
--- a/packages/plugins/google-secret-manager/README.md
+++ b/packages/plugins/google-secret-manager/README.md
@@ -16,7 +16,7 @@ See [our docs site](https://varlock.dev/plugins/google-secret-manager/) for comp
 # @initGsm(projectId=my-gcp-project, credentials=$GCP_SA_KEY)
 # ---
 
-# Service account JSON key (for production/CI)
+# Service account JSON key (optional)
 # @sensitive @type=gcpServiceAccountJson
 GCP_SA_KEY=
 
--- a/packages/plugins/google-secret-manager/package.json
+++ b/packages/plugins/google-secret-manager/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@varlock/google-secret-manager-plugin",
   "description": "Varlock plugin to pull data from Google Cloud Secret Manager",
-  "version": "0.1.0",
+  "version": "0.0.0",
   "type": "module",
   "repository": {
     "type": "git",


philmillman:
Thanks for this @ya7010, we'll take a look soon. cc: @theoephraim

packages/plugins/google-secret-manager/src/plugin.ts
```diff
@@ -0,0 +1,357 @@
...
+        const client = new SecretManagerServiceClient(clientConfig);
+        debug('GSM client initialized for instance', this.id);
+        return client;
+      } catch (err) {
+        const errorMsg = err instanceof Error ? err.message : String(err);
+        throw new SchemaError(`Failed to initialize Google Secret Manager client: ${errorMsg}`, {
+          tip: [
+            'Verify service account JSON is valid',
+            'For ADC: Run `gcloud auth application-default login`',
+            'Check GOOGLE_APPLICATION_CREDENTIALS environment variable',
```
theoephraim, line 73: Not sure about this - is this something that gets injected automatically?

packages/plugins/google-secret-manager/src/plugin.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    return `projects/${this.projectId}/secrets/${name}/versions/${version}`;
+  }
+
+  readBatch?: Record<string, { defers: Array<DeferredPromise<string>> }> | undefined;
+
+  async readSecret(secretRef: string): Promise<string> {
+    await this.initClient();
+
+    // Simple batching setup to parallelize reads
```
theoephraim: I assume this was all copied / adapted from 1Password, where batching reads is important. In this case, without any special batch endpoints, I think we can just simplify this and read the secrets directly without batching?

packages/plugins/google-secret-manager/src/plugin.ts
```diff
@@ -0,0 +1,357 @@
...
+    }
+    pluginInstances[id] = new GsmPluginInstance(id);
+
+    if (objArgs.projectId && !objArgs.projectId.isStatic) {
+      throw new SchemaError('Expected projectId to be static');
+    }
+    const projectId = objArgs?.projectId ? String(objArgs.projectId.staticValue) : undefined;
+
+    // User should either be setting credentials, useADC, or both
+    if (!objArgs.credentials && !objArgs.useADC) {
```
theoephraim: We maybe could just default to using ADC if no credentials are set, since ADC only works in deployed environments? For 1pass its more important to let the user opt in, since it accesses your local 1pass app, which often has way more privileges.

What do you think?

packages/plugins/google-secret-manager/tsup.config.ts
```diff
@@ -0,0 +1,27 @@
...
+  dts: true,
+  sourcemap: true,
+  treeshake: true,
+  clean: false,
+  outDir: 'dist',
+  format: ['esm'],
+  splitting: false,
+  target: 'esnext',
+  external: ['varlock'],
+  banner: ({ format }) => {
```
theoephraim, line 14: You likely don't need this part - it is just for the 1pass sdks which have some cjs/esm issues... Test it without and see if it works.

theoephraim (commented):
Sorry for the delay here - just getting back online after the holidays.

This is awesome - thanks so much! Just a couple small notes, and we can get this merged asap :)

packages/plugins/google-secret-manager/tsup.config.ts
ya7010: After deleting this, it stopped working.

```
$ node packages/varlock/bin/cli.js load -f json 

🚨 🚨 🚨  Configuration is currently invalid   🚨 🚨 🚨

Invalid items:

❓ GCP_SECRET_TEST_KEY*  🔐sensitive
   └ undefined
   - Value is required but is currently empty


💥 Resolved config/env did not pass validation 💥
```

packages/plugins/google-secret-manager/src/plugin.ts
ya7010: That's good!

I've modified it to use ADC when credentials are not specified.

https://github.com/dmno-dev/varlock/pull/237/changes/7ae544ccbe527b1a5380499829b84695207ec246

packages/plugins/google-secret-manager/src/plugin.ts
ya7010: fix it.

https://github.com/dmno-dev/varlock/pull/237/changes/2e099717c5e83cad0c5a7c3b5b8ad2ed74113da3

packages/plugins/google-secret-manager/src/plugin.ts
ya7010: Regarding ADC, it will be automatically inserted.

theoephraim (approved):

theoephraim:
@ya7010 - I've added some docs and will merge and publish. Please take a look at the [new docs page](https://varlock.dev/plugins/google-secret-manager/) and open a new PR if you notice any issues. I'm not a GCP / SM user, so your eyes here would be much appreciated!

Issue #179, opened 2025-10-29
[BUG]: better handling of multiple `@generateTypes()` calls


theoephraim:
### varlock version

0.0.15

### Steps to reproduce

Import multiple `.env.schema` files which contain `@generateTypes` calls.

### What is expected?

Only the primary file will have its types generated, including imported items. Type generation in the imported services will be skipped.

### What is actually happening?

Each service has its types generated, but includes all items.

### Any additional comments?

Currently the behaviour is that type generation is skipped if the file is imported _partially_ (meaning specific keys are imported). The intention here is that you might just be breaking up a large schema, so we dont necessarily know in what cases you wanted to skip type generation in an imported file in all cases.

One possible solution is trying to infer if the file is the _main_ schema, or doing some special handling of files named `.env.schema`.

I believe a better solution is to let the user trigger this _partial import_ treatment, even when importing all items. You're letting the user say - "import this file fully as if it was part of this one" versus "import all items from this other file".

This could be implemented by adding wildcards to the import list. For example
- `@import(../other-file)` - imports fully, a contained `@generateTypes` call would be executed
- `@import(../other-file, *)` - imports partially, all items included, but a contained `@generateTypes` call is skipped

theoephraim:
Ok I'm rethinking this a bit, and now I'm thinking that `@generateTypes()` will be ignored whenever it is present within an imported file. I think this should reflect the user intent in most cases.

The only case where I can imagine you might not want this is if say you had your main `.env.schema` only contained an import of some config folder, just because you wanted to organize your .env files in a folder. In this case, we can add a `executeWhenImported` flag on the `@generateTypes` function itself.

PR #245, opened 2026-01-06
disable generateTypes in imported files

theoephraim:
addresses #179 

This disables type generation in any imported file - which I believe will match user intent in most cases.
If the user needs to, they can explicitly override that with  a new flag `@generateTypes(executeWhenImported=true)`. 

The alternative is to let the user specifically opt in to a "partial import" by using some kind of regex or boolean flag. However currently this concept is not used for anything else, so I think it may be better to hide it.

The downside here is that other decorators in the future may need similar handling, but so far it seems that the behaviour is somewhat custom to the decorator itself, and is not common.

### DIFF
--- a/packages/varlock-website/src/content/docs/guides/import.mdx
+++ b/packages/varlock-website/src/content/docs/guides/import.mdx
@@ -67,7 +67,6 @@ A registry (npm, jsr, etc.), package name, version, and path can be used to impo
 By default, all items will be imported, but you may add a list of specific keys to import as additional args after the first.
 
 - If there is a chain of imports, an item is only imported if every ancestor import includes it (or imports all items)
-- If a file is partially imported, some root decorators that affect the overall system behavior rather than individual items will be ignored (e.g., `@generateTypes`)
 
 ```env-spec
 # @import(./.env.imported, KEY1, KEY2)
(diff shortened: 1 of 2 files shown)


packages/varlock/src/env-graph/lib/data-source.ts
```diff
@@ -57,7 +57,7 @@ export abstract class EnvGraphDataSource {
     return !!this.importMeta?.isImport || !!this.parent?.isImport;
   }
   get isPartialImport() {
-    return (this.importMeta?.importKeys || []).length > 0;
+    return (this.importKeys || []).length > 0;
```
theoephraim, line 60: this logic was wrong - because a parent could be a partial import

packages/varlock/src/env-graph/lib/env-graph.ts
```diff
@@ -340,8 +340,6 @@ export class EnvGraph {
     const sources = Array.from(this.sortedDataSources).reverse();
     for (const s of sources) {
       if (s.disabled) continue;
-      // we skip root decorators if the file was being _partially_ imported
-      if (s.isPartialImport) continue;
```
theoephraim, line 344: for now, we'll let individual decorators deal with this.. most decorators are used on a single source at a time anyway

PR #250, opened 2026-01-07
fix docs related to .env file precedence order

### LINKED ISSUE #249: .env.local does not override .env.[env] files
## Description

`.env.local` does not override environment-specific files like `.env.production` or `.env.development`, contrary to the documented behavior.

## Reproduction

Minimal repro: https://github.com/kasperpeulen/varlock-env-local-repro

```bash
git clone https://github.com/kasperpeulen/varlock-env-local-repro
cd varlock-env-local-repro
bun install
APP_ENV=production npx varlock load
```

**Expected:** `PORT=3000` (from `.env.local`)  
**Actual:** `PORT=8081` (from `.env.production`)

## Documentation vs Reality

According to [the docs](https://varlock.dev/guides/environments/), the loading order (lowest to highest priority) should be:

1. `.env.schema`
2. `.env.[currentEnv]`
3. `.env.local` ← should override
4. `.env.[currentEnv].local`

But `varlock load --format json-full` shows sources in this order:

```json
"sources": [
  { "label": ".env.production", "enabled": true },
  { "label": ".env.local", "enabled": true },
  { "label": ".env.schema", "enabled": true }
]
```

This suggests `.env.production` is applied AFTER `.env.local`, effectively overriding it instead of the other way around.

## Workaround

Using `.env.[env].local` files works correctly:

```bash
echo "PORT=9999" > .env.production.local
APP_ENV=production npx varlock load
# PORT=9999 ✅
```

## Environment

- varlock: 0.1.4
- bun: 1.3.5
- node: v22.x
- macOS Darwin 24.6.0

theoephraim:
fixes #249 - updates docs to match reality, which is that `.env.local` has less precedence than  `.env.[currentEnv]`


philmillman (approved):
:shipit:

PR #262, opened 2026-01-16
feat: add `env-compact` format.

ya7010:
Currently, the `env` format of the `load` command passes an empty string for undefined environment variables.
This can cause applications to recognize the environment variable as existing.

As a countermeasure, I have added `env-compact` format that does not output undefined environment variables.

I considered adding `--skip-undefined` option, but since it would only affect `env`, so I have added a new format instead.

### DIFF
--- a/.changeset/honest-boats-attend.md
+++ b/.changeset/honest-boats-attend.md
@@ -1,0 +1,5 @@
+---
+"varlock": patch
+---
+
+feat: add `--compact` flag `varlock load`.


ya7010:
@theoephraim 

I'm torn between defining a new format or just adding more options — which do you think is better?

theoephraim:
@ya7010 - I think an option makes sense. The compact flag still kind of makes sense in other formats. For example in the pretty printed format, it could just skip empty items.

ya7010:
@theoephraim 

Added the `--compact` option and modified it to skip undefined environment variables when used with the `env` format.

theoephraim (approved):

Issue #217, opened 2025-11-25
[Feature Request] Multi-line @import statements


benvinegar:
### Description

I have @import statements that are like 30 different variables (sigh).

It would be much easier / more readable if I could put them on separate lines somehow.

```
# @import(../../.env.schema, EXAMPLE_A, EXAMPLE_B, EXAMPLE_C, EXAMPLE_D, ... EXAMPLE_Z)
```

theoephraim:
Multiline function args is definitely on our wishlist but it will require some bigger changes to the parser itself.

In the meantime, would regex matching be helpful?
`# @import(../.env.other, regex("EXAMPLE_.*"))`

Or using some kind of tags?
`# @import(../.env.other, tag=featurex)`

In other file, tagging the specific items:
```
# @tag(featurex)
FEATURE_X_KEY=
```

benvinegar:
Yeah maybe regex would help a bunch. Also I think I can clean up how many imports I have, so ... unclear how important this is to me just yet.

PR #278, opened 2026-02-06
add multi-line fn calls

theoephraim:
Adds multi-line support for function calls, both in decorators and item values.

Big pain that this solves is very long lists of imported keys. see #217 

Supporting this in item values does break our previous goal of not breaking other .env parsers, but I think it may be worthwhile. If a specific user needs to maintain support for another env loader for some reason, they can just choose to not use multi-line fn call values. This should be relatively rare unless you start getting into very complicated composition of values and functions...

Follow up work here is
- support nested function calls within decorators
- support commenting out whole lines within a multi line fn call
- support trailing comments after a line within a multi-line fn call
- support same commenting within decorator multi-line fns

### DIFF
--- a/packages/varlock/src/env-graph/lib/data-source.ts
+++ b/packages/varlock/src/env-graph/lib/data-source.ts
@@ -4,7 +4,7 @@ import path from 'node:path';
 import _ from '@env-spec/utils/my-dash';
 import { tryCatch } from '@env-spec/utils/try-catch';
 import {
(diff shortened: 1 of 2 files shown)


.changeset/beige-impalas-talk.md
```diff
@@ -0,0 +1,6 @@
+---
+"@env-spec/parser": patch
+"varlock": patch
```
philmillman: We might consider a minor version here since it's potentially breaking and files with multiline values won't work in older versions

theoephraim: good catch!

PR #283, opened 2026-02-10
new --path/-p flag to load from a specific path rather than assuming current directory

theoephraim:
Adds a new `--path` / `-p` flag to `load` and `run` commands to specify a .env file or directory as the entry point


philmillman (approved):
lgtm :shipit:

PR #292, opened 2026-02-12
aws plugin initial version

theoephraim:


packages/plugins/aws-secrets/src/plugin.ts
```diff
@@ -0,0 +1,716 @@
...
+        // Check if we're using explicit credentials or default chain
+        if (!this.accessKeyId && !this.profile) {
+          errorMessage = 'Authentication failed';
+          errorTip = [
+            err.message,
+            'To fix this, choose one of the following:',
+            '  1. Set AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables',
+            '  2. Configure ~/.aws/credentials file (run: aws configure)',
+            '  3. Provide credentials explicitly via @initAws(accessKeyId=..., secretAccessKey=...)',
+            '  4. Use IAM roles (if running on AWS infrastructure)',
```
philmillman: looks like we use this or something like it in a few places, it's probably a good idea to have a single version so it's easier to maintain

PR #306, opened 2026-02-17
fix windows smoke tests issue

theoephraim:


philmillman (approved):
Lgtm! might eventually want those timeouts in a const/config item if they're being used in more than a couple places

PR #313, opened 2026-02-24
use-bun

### LINKED ISSUE #312: [BUG] `PKG_EXECPATH` from parent varlock process breaks nested `varlock run` calls
### varlock version

0.1.4

### Steps to reproduce

1. Launch a process via varlock:
   ```bash
   varlock run -- node supervisor.js
   ```
2. Inside `supervisor.js` (or a shell script spawned by it), call varlock again:
   ```bash
   varlock run --path ~/.config/ -- sh -c 'echo $MY_VAR'
   ```
3. The nested `varlock run` fails:
   ```
   Error: Cannot find module '/home/user/run'
   ```

### What is expected?

Nested `varlock run` calls should work correctly regardless of how the parent process was launched.

The initial `varlock run` sets `PKG_EXECPATH=/path/to/varlock` in the child process environment. This is a Node.js SEA (Single Executable Application) convention — when set, it tells Node.js that the binary was packaged as a single executable.

When a nested `varlock run` is invoked, the varlock binary (which is a Node.js SEA) sees `PKG_EXECPATH` already set and misinterprets its own `argv`. Instead of treating `run` as a subcommand, it tries to resolve it as a Node module path relative to the working directory, producing:

```
Error: Cannot find module '/home/user/run'
```

### Workaround

Unset `PKG_EXECPATH` before any nested varlock call:

```bash
unset PKG_EXECPATH
varlock run --path ~/.config/ -- node worker.js
```

### Suggested fix

varlock should `unset PKG_EXECPATH` (or delete it from `process.env`) early in its own startup, before argument parsing. The binary knows it is a SEA — it does not need this env var to persist, and it should not leak it to child processes.

Alternatively, varlock could strip `PKG_EXECPATH` from the child process environment in `varlock run` so it does not propagate.

### System Info

- OS: Ubuntu 24.04 (aarch64)
- Node: v22.14.0 (embedded in varlock SEA)
- varlock: 0.1.4
- Install: `~/.varlock/bin/varlock`

### Context

Discovered in a long-running supervisor process launched via `varlock run -- node supervisor.js`. When the supervisor spawns helper scripts that also invoke `varlock run` to probe for configured secrets, all nested varlock calls fail silently (stderr is suppressed with `2>/dev/null` in production), causing downstream processes to never start.

theoephraim:
- moves over from pnpm to bun as our package manager
- moves from pkg and node-powered SEA to a bun powered standalone binary

Fixes #312 - and added some basic tests around the issue.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,8 +19,8 @@
 ## Binary builds
 
 - The varlock CLI binary is built using `bun build --compile` (not Node SEA or pkg)
-- `bun run --filter varlock build:sea` builds a local dev binary at `packages/varlock/dist-sea/varlock`
-- `packages/varlock/scripts/build-binaries.ts` builds cross-platform release binaries
+- `bun run --filter varlock build:binary` builds a local dev binary for the current platform at `packages/varlock/dist-sea/varlock`
+- `packages/varlock/scripts/build-binaries.ts` builds cross-platform release binaries (or use `--current-platform` for a single local binary)
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -18,10 +18,11 @@
 
 - 🛡️ validation, coercion, type safety w/ IntelliSense
 - 🔏 protection for sensitive config values (log redaction, leak prevention)
-- 🌐 flexible multi-environment management
-- 💫 composition of values, functions, load from external sources
+- 🌐 flexible multi-environment management - auto .env.* loading and explicit import
+- 💫 flexible composition and toggling using functions() and $references
+- 🔌 plugin system for secure secret loading from various sources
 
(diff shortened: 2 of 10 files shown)


theoephraim:
cloudflare failures are due to config settings that live within the cloudflare UI calling pnpm - we will update these when ready to merge

.changeset/twelve-beers-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"varlock": patch
```
philmillman: I think this needs to be, at minimum, a minor version

philmillman:
We'll need to update the build commands in cloudflare for the website and the docs-mcp or move them into config files

PR #318, opened 2026-02-25
use pnpm for smoke tests

theoephraim:
smoke tests are failing because bun isnt as flexible in dealing with locally linked packages, and installing from tgz files, particularly with overrides. So we are swapping back to pnpm just for the smoke tests. Ideally we'd have another set of tests just checking each package manager, particularly for the nextjs overrides stuff, but even that might be tough, because the overrides dont work with local files in all of them.

Also our new binary tests caught an issue with the windows bun binary, so they are working :)

Much deeper to go here, but even this first pass is a great start.


philmillman (approved):
🚢

PR #342, opened 2026-03-09
consistent type generation

theoephraim:
- Move type generation before resolveEnvValues() for deterministic output
- Track isEnvSpecific on data sources (applyForEnv, conditional @disable/@import)
- Filter env-specific sources from type gen via defsForTypeGeneration
- Skip items only defined in env-specific files from generated types
- Add auto=false param to @generateTypes to disable during load/run
- Add varlock typegen command for manual type generation
- 25 new tests for type generation scenarios


theoephraim:
fixes #335 and #263

PR #352, opened 2026-03-10
feat: support $XDG_CONFIG_HOME for user config directory

### LINKED ISSUE #348: Please support $XDG_CONFIG_HOME
[XDG Base Directory Spec](https://specifications.freedesktop.org/basedir/latest/#:~:text=should%20be%20used.-,%24XDG_CONFIG_HOME,/.config%20should%20be%20used.,-%24XDG_STATE_HOME)

This would need that [this line](https://github.com/dmno-dev/varlock/blob/c4270598d8c61af3dbc7975577cda6f7c3858826/packages/varlock/src/cli/helpers/telemetry.ts#L25) defaults to XDG_CONFIG_HOME instead of os.get_homedir.

The spec says that you should default to `~/.config/` if it's unset.

For backwards compatibility, you could check if `~/.varlock` exists and use it, else use `~/.config/varlock`.
Alternatively, although technically not spec-compliant, swap that check (several other projects have gone this approach).

theoephraim:
Closes #348

Adds support for the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/) for the user-level varlock config directory.

### Resolution order
1. If `$XDG_CONFIG_HOME` is set → `$XDG_CONFIG_HOME/varlock`
2. If legacy `~/.varlock` exists → `~/.varlock` (backwards compatibility)
3. Otherwise → `~/.config/varlock` (XDG default)

### Changes
- **New**: `packages/varlock/src/lib/user-config-dir.ts` — shared utility for resolving the user config dir
- **Updated**: `telemetry.ts`, `telemetry.command.ts`, `plugins.ts` — use the new utility instead of hardcoded `~/.varlock`
- **Updated**: `install.sh` — mirrors the same XDG resolution logic in shell
- **Updated**: docs and error messages to reference the new paths
- **Updated**: `login.command.ts` — updated TODO comments

### DIFF
--- a/.changeset/warm-keys-play.md
+++ b/.changeset/warm-keys-play.md
@@ -1,0 +1,5 @@
+---
+"varlock": patch
+---
+
+Support XDG Base Directory Specification for user config directory. Varlock now respects `$XDG_CONFIG_HOME` and defaults to `~/.config/varlock` instead of `~/.varlock` for new installations, while maintaining backwards compatibility with existing `~/.varlock` directories.
--- a/packages/varlock-website/src/content/docs/guides/telemetry.mdx
+++ b/packages/varlock-website/src/content/docs/guides/telemetry.mdx
@@ -28,7 +28,7 @@ Run the following command to permanently opt out:
 
 <ExecCommandWidget command="varlock telemetry disable" />
 
-This will create/update a configuration file saving your preference at `~/.varlock/config.json`.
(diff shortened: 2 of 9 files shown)


theoephraim:
@copilot - lets check how the plugin cache interacts with this. Specifically in the case where the user has no home folder, like in docker. Its ok if the config isnt there because we can skip telemetry, but the plugin cache needs to be written somewhere. It could be in the current folder, or in /tmp - although within docker the user may need to copy those files to another layer. Figure out what makes the most sense for this docker-with-no-home-folder case

PR #407, opened 2026-03-13
Fix Bitwarden parallel auth requests from triggering rate limits

benperove:
## Problem

When the `bitwarden()` resolver fetches multiple secrets, Varlock resolves them in **parallel**. Each secret fetch calls `authenticate()` to obtain a JWT from Bitwarden's identity server. Because they run concurrently:

- All calls see `!this.cachedAuth` simultaneously
- Each starts its own request to `identity.bitwarden.com/connect/token`
- Result: N auth requests instead of 1 (e.g. 7 secrets → 7 auth calls)
- Bitwarden's rate limit is hit immediately → HTTP 429

This makes the plugin unusable for schemas that load multiple secrets (typical setup).

## Solution

Add an in-flight lock so concurrent `authenticate()` calls share a single auth request. The first call performs the auth; subsequent callers await the same promise and reuse the result.

## Changes

- Add `private authInFlight?: Promise<CachedAuth>` to `BitwardenPluginInstance`
- Refactor `authenticate()`: when auth is in progress, await it; otherwise start a new auth and set `authInFlight`
- Extract the actual token exchange logic into `_doAuthenticate()` for clarity

## Testing

- Verified locally with a schema loading 7 secrets from Bitwarden
- Before: 429 rate limit on first run
- After: Single auth request, all secrets load successfully

### DIFF
--- a/packages/plugins/bitwarden/src/plugin.ts
+++ b/packages/plugins/bitwarden/src/plugin.ts
@@ -96,6 +96,8 @@ class BitwardenPluginInstance {
 
   /**
    * Authenticate and get JWT + organization key
+   * Uses authInFlight to prevent parallel secret resolution
+   * from triggering multiple auth requests & getting rate limited
    */
   private async authenticate(): Promise<CachedAuth> {
     if (!this.accessToken) {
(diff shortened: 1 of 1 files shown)


theoephraim:
Awesome. I did some light testing of Bitwarden but don't use it day to day so missed this obvious problem.

Thanks so much for the PR!

theoephraim:
@benperove - looks like some of your changes are missing!

benperove:
@theoephraim Sorry about that. I added another commit.

theoephraim (approved):

PR #467, opened 2026-03-23
matej/vscode-diagnostics-comment-awareness

voiys:
## Summary
- only treat comment lines as decorator-bearing when the comment body starts with `@`
- stop regular comments that merely mention `@decorator` or `@type=...` from triggering VS Code diagnostics/type validation
- add regression coverage for both regular-comment cases

## Test plan
- [x] `bunx vitest packages/vscode-plugin/test/diagnostics-core.test.ts`
- [x] `bun run lint:fix`

Examples:

1. Regular comments are ignored (intellisense does not trigger either, can turn that off if you want)
Screenshot 2026-03-23 at 5 50 32 PM

2. Post-comments on decorator lines are ignored
Screenshot 2026-03-23 at 5 46 19 PM

3. Real decorator diagnostics still work
Screenshot 2026-03-23 at 5 46 28 PM

4. Real type diagnostics still work, but commented trailing type text does not
Screenshot 2026-03-23 at 5 46 42 PM
Screenshot 2026-03-23 at 5 46 49 PM

### DIFF
--- a/.changeset/matej-vscode-diagnostics-comment-awareness.md
+++ b/.changeset/matej-vscode-diagnostics-comment-awareness.md
@@ -1,0 +1,5 @@
+---
+"env-spec-language": patch
+---
+
+Fix VS Code diagnostics and completions so decorator parsing ignores prose mentions and post-comments while still matching parser behavior for leading `@word` comment lines.
--- a/packages/vscode-plugin/test/diagnostics-core.test.ts
+++ b/packages/vscode-plugin/test/diagnostics-core.test.ts
@@ -40,6 +40,43 @@ describe('diagnostics-core', () => {
     );
   });
 
+  it('ignores decorator-like text inside regular comments', () => {
+    const diagnostics = createDecoratorDiagnostics(
+      getDecoratorOccurrences('# this @required mention is just documentation', 0),
+    );
+
+    expect(diagnostics).toEqual([]);
+  });
+
+  it('ignores decorator-like text inside post-comments on decorator lines', () => {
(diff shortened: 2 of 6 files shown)


theoephraim (approved):

packages/vscode-plugin/test/diagnostics-core.test.ts
```diff
@@ -36,6 +36,27 @@ describe('diagnostics-core', () => {
...
+
+  it('ignores decorator-like text inside post-comments on decorator lines', () => {
+    const diagnostics = createDecoratorDiagnostics(
+      getDecoratorOccurrences('# @required # this @optional is commented', 0),
+    );
+
+    expect(diagnostics).toEqual([]);
+  });
+
+  it('ignores parser-style ignored decorator comments', () => {
```
theoephraim: Not sure about this. The real loader will explode when it sees this with an "unknown decorator" kind of error. Maybe we should allow them as noops since they are common though. But either way it should match.

.changeset/matej-vscode-diagnostics-comment-awareness.md
```diff
@@ -0,0 +1,2 @@
+---
```
theoephraim, line 1: you can add a changeset with the vscode extension tagged. It will merge them all up when we publish together

voiys:
VS Code now matches parser behavior here: prose mentions/post-comments are ignored, but leading `@word` comment lines still parse the same way the parser does

Screenshot 2026-03-25 at 11 52 40 PM

PR #475, opened 2026-03-23
Add ts-nocheck directive to type-generation.ts

developerzeke:
I keep accidentally opening this autogenerated file, which causes my IDE to try to typecheck it, and it fails with a `TS2344: Type '' does not satisfy the constraint 'DATABASE_URL'` error. It seems to work fine, so this would silence the error entirely by adding a `@ts-nocheck` comment to the top of the file.

### DIFF
--- a/.changeset/lazy-paws-decide.md
+++ b/.changeset/lazy-paws-decide.md
@@ -1,0 +1,5 @@
+---
+"varlock": patch
+---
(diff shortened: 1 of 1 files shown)


theoephraim:
makes sense!

theoephraim (approved):

Issue #367, opened 2026-03-11
[Feature Request] An official plugin for Proton password manager


gormus:
### Description

An official plugin that enables secure loading of values from Proton Password Manager.

Proton Pass
https://proton.me/pass

philmillman:
It appears they only have a recently introduced a CLI, and it's a paid feature: https://proton.me/blog/proton-pass-cli. We're happy to support this, but just wanted to highlight this requirement as it might exclude a lot of users.

developerzeke:
this would be great, I use Proton for my normal password manager and would prefer to have it all in one place. I have premium so that wouldn't be an issue for me

philmillman:
> this would be great, I use Proton for my normal password manager and would prefer to have it all in one place. I have premium so that wouldn't be an issue for me

Ok if you don't mind a bit of beta testing we can probably get something out this week!

philmillman:
Preview packages are here if anyone wants to test this: https://github.com/dmno-dev/varlock/pull/487, I will test this afternoon but that requires a paid account so if you already have one then testing is much appreciated!!

developerzeke:
I'm having trouble getting it installed :/ I tried putting it in the root monorepo package and the individual apps but neither worked. it's a file: url to a clone of the varlock repo since it's not published yet. I the `npm i https://pkg.pr.new/dmno-dev/varlock/@varlock/proton-pass-plugin@487` command sent by the bot, but it gave me an error about using `workspace:` in the package.json of that package. Does the preview only work with Bun?

philmillman:
Hmm I'm not sure, that might be a symptom of the new publishing workflow for plugins, @theoephraim any thoughts?

theoephraim:
Fix incoming! When we moved from pnpm to bun, we had to swap these catalog/workspace deps ourselves, and we forgot to apply it to preview packages.

philmillman:
fix is in that PR now

developerzeke:
It mostly works! Initializing without an ID doesn't work:

```
# @plugin(@varlock/proton-pass-plugin)
# @initProtonPass()
# @generateTypes(lang=ts, path=env.d.ts)
# ----------

# Connection string to a Postgres database
# @required @sensitive
DATABASE_URL=protonPass(pass://env/myapp/db_url)
```


```
🚨 Error(s) encountered in .env.schema
- Expected configuration arguments


🚨 🚨 🚨  Configuration is currently invalid   🚨 🚨 🚨

[...]

PR #487, opened 2026-03-25
proton pass plugin

philmillman:
implements #367

### DIFF
--- a/packages/plugins/proton-pass/src/plugin.ts
+++ b/packages/plugins/proton-pass/src/plugin.ts
@@ -3,7 +3,7 @@ import { ExecError, spawnAsync } from '@env-spec/utils/exec-helpers';
 
 const { ValidationError, SchemaError, ResolutionError } = plugin.ERRORS;
 
-const PROTON_PASS_ICON = 'simple-icons:protonpass';
+const PROTON_PASS_ICON = 'simple-icons:proton';
 
 plugin.name = 'proton-pass';
 const { debug } = plugin;
--- a/packages/varlock-website/src/content/docs/plugins/proton-pass.mdx
+++ b/packages/varlock-website/src/content/docs/plugins/proton-pass.mdx
@@ -72,6 +72,8 @@ The password/TOTP/extra password are provided to the CLI via environment variabl
 
 :::tip[Already logged in?]
 If `pass-cli` is already logged in, you can also omit credentials and just rely on the existing session.
+
+This mostly applies to local development, in CI, you will probably want static credentials.
 :::
 
 ## Loading secrets


packages/plugins/proton-pass/src/plugin.ts
```diff
@@ -0,0 +1,491 @@
+import { Resolver } from 'varlock/plugin-lib';
+import { ExecError, spawnAsync } from '@env-spec/utils/exec-helpers';
+
+const { ValidationError, SchemaError, ResolutionError } = plugin.ERRORS;
+
+const PROTON_PASS_ICON = 'simple-icons:protonpass';
```
theoephraim: this is hallucinated. Might need to provide a full SVG here. I'm not sure if its set up to accept that yet or not...

packages/plugins/proton-pass/src/plugin.ts
```diff
@@ -0,0 +1,491 @@
...
+plugin.name = 'proton-pass';
+const { debug } = plugin;
+debug('init - version =', plugin.version);
+plugin.icon = PROTON_PASS_ICON;
+
+plugin.standardVars = {
+  initDecorator: '@initProtonPass',
+  params: {
+    password: { key: 'PROTON_PASS_PASSWORD', dataType: 'protonPassPassword' },
+    totp: { key: 'PROTON_PASS_TOTP', dataType: 'protonPassTotp' },
```
theoephraim, line 17: how does this work? "totp" usualyl means time bounded, so wouldnt really make sense to pass in this way.

packages/varlock-website/src/content/docs/plugins/proton-pass.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+# @type=protonPassExtraPassword @sensitive
+PROTON_PASS_EXTRA_PASSWORD=
+```
+
+The `username` is passed as an argument to `pass-cli login --interactive <username>`.
+The password/TOTP/extra password are provided to the CLI via environment variables, as supported by Proton Pass CLI login docs:
+[login command](https://protonpass.github.io/pass-cli/commands/login/).
+
+:::tip[Already logged in?]
```
theoephraim, line 73: might want to clarify that this only makes sense for local dev

packages/plugins/proton-pass/src/plugin.ts
philmillman: yeah I guess we could do it magically for you if you're using 1pass (or another provider) but I doubt most proton users are using another managed service for totp. 

their docs seem to imply a static totp though: https://protonpass.github.io/pass-cli/commands/login/#password

packages/varlock-website/src/content/docs/plugins/proton-pass.mdx
philmillman: yeah will do

packages/plugins/proton-pass/src/plugin.ts
philmillman: I'll just make it generic for now

PR #496, opened 2026-03-26
fix: @initAws profile graceful fallback to default credential chain

SeeThruHead:
When `@initAws(profile="...")` is specified but the profile is unavailable (no `~/.aws/config`, missing profile entry), the plugin now falls back to the SDK's default credential chain instead of hard-failing.

**Problem:** The same `.env.schema` can't work in both local dev (SSO profiles in `~/.aws/config`) and containers/k8s (IRSA or env var credentials). In containers there's no config file, so `fromIni({ profile })` throws and the entire load fails — even though valid credentials are available via env vars or IRSA.

**Fix:** Eagerly test the profile provider. If it resolves, use it. If it throws, skip it and let the SDK use its default credential chain (env vars → IRSA → instance metadata).

**Behavior change:**
- Profile exists → used as before (no change)
- Profile missing → falls back to default chain (previously crashed)
- No profile specified → default chain as before (no change)
- Explicit credentials → used as before (no change)

### DIFF
--- a/.changeset/aws-profile-fallback.md
+++ b/.changeset/aws-profile-fallback.md
@@ -1,0 +1,5 @@
+---
+"@varlock/aws-secrets-plugin": patch
+---
+
+Fall back to default AWS credential chain when @initAws profile is unavailable (e.g. no ~/.aws/config in containers/k8s). Previously this would hard-fail even when valid credentials were available via env vars or IRSA.


theoephraim:
Wrote this note on discord too, but will leave here in case anyone else is following along

---

I'm not sure I love the idea of it gracefully falling back to the other method. If you are trying to use a profile and it doesn't exist, it could potentially hide the error and do something you weren't expecting. What about using functions to enable the profile in the correct scenario only.

So for example `@initAws(region=us-east-1, profile=if(forEnv(production), profileName))`
or `@initAws(region=us-east-1, profile=if($IS_DOCKER, profileName))`

You could even do
`@initAws(region=us-east-1, profile=if(isEmpty($AWS_ACCESS_KEY_ID), profileName))`
if you wanted to enable the profile only if the env var is not set


Any reason that this wouldn't make sense or be doable?

If this was behaviour that the aws cli did natively and folks were expecting, then maybe, but from what I can tell, the aws cli  would fail if you passed in a profile and it doesn't exist.

PR #497, opened 2026-03-27
Fix plugin export path for HashiCorp Vault

benperove:
# Fix plugin export path for HashiCorp Vault

## What was wrong

The `package.json` was trying to load `dist/plugin.js`, but the build creates `dist/plugin.cjs`. Varlock looks up the path and couldn’t find the file and it fails to load.

## Fix

Point the `./plugin` export at `dist/plugin.cjs` so it matches the build output (same pattern as the Bitwarden plugin).

## How to verify

- Build `packages/plugins/hashicorp-vault` and confirm `dist/plugin.cjs` is present.
- In a project using `@plugin(@varlock/hashicorp-vault-plugin)` and `@initHcpVault(...)`, Varlock should load the plugin without a “plugin file not found” error.

### DIFF
--- a/.changeset/chilly-cats-know.md
+++ b/.changeset/chilly-cats-know.md
@@ -1,0 +1,5 @@
+---
+"@varlock/hashicorp-vault-plugin": patch
+---
+
+Fix plugin export path for HashiCorp Vault


theoephraim:
Sorry about that... doing lots of things at once :)

Thanks for the PR!

PR #498, opened 2026-03-27
* feat: add Passbolt plugin

PaddeK:
- Adds @varlock/passbolt-plugin. Registers @initPassbolt decorator, passbolt()/passboltFolder()/passboltCustomFields() resolvers, and passboltAccountKit data type.
- Add comprehensive plugin README with usage examples
- Update root README plugin table

### DIFF
--- a/packages/varlock-website/astro.config.ts
+++ b/packages/varlock-website/astro.config.ts
@@ -189,6 +189,7 @@ export default defineConfig({
             { label: 'Infisical', slug: 'plugins/infisical' },
             { label: 'KeePass', slug: 'plugins/keepass' },
             { label: 'Pass', slug: 'plugins/pass' },
+            { label: 'Passbolt', slug: 'plugins/passbolt' },
             { label: 'Proton Pass', slug: 'plugins/proton-pass' },
           ],
         },
--- a/packages/varlock-website/src/content/docs/plugins/overview.mdx
+++ b/packages/varlock-website/src/content/docs/plugins/overview.mdx
@@ -22,6 +22,7 @@ For now, only official Varlock plugins under the `@varlock` npm scope are suppor
 | [Infisical](/plugins/infisical/) | <Badge npmPackage="@varlock/infisical-plugin" /> |
 | [KeePass](/plugins/keepass/)<br/>_KDBX 4.0 databases & KeePassXC CLI_ | <Badge npmPackage="@varlock/keepass-plugin" /> |
 | [Pass](/plugins/pass/)<br/>_the standard unix password manager_ | <Badge npmPackage="@varlock/pass-plugin" /> |
+| [Passbolt](/plugins/passbolt/)<br/>_open-source password manager for teams_ | <Badge npmPackage="@varlock/passbolt-plugin" /> |
 | [Proton Pass](/plugins/proton-pass/)<br/>_CLI-backed Proton password manager_ | <Badge npmPackage="@varlock/proton-pass-plugin" /> |
 
 
--- a/packages/varlock-website/src/content/docs/plugins/passbolt.mdx
+++ b/packages/varlock-website/src/content/docs/plugins/passbolt.mdx
@@ -1,0 +1,308 @@
+---
+title: Passbolt Plugin
+description: Using Passbolt with Varlock
+---
+
+import { Steps, Icon } from '@astrojs/starlight/components';
+import Badge from '@/components/Badge.astro';
+
+<div class="page-badges">
+  <Badge npmPackage="@varlock/passbolt-plugin" />
+</div>
+
+Our [Passbolt](https://www.passbolt.com/) plugin enables secure loading of secrets from Passbolt password manager using declarative instructions within your `.env` files.
+
+Passbolt is an open-source, self-hosted password manager built for teams. This plugin connects to your Passbolt instance's API using your account kit and passphrase, decrypting secrets via OpenPGP.
+
+## Features
+
+- **Account kit authentication** — just provide your account kit and passphrase
+- **UUID-based secret access** — fetch secrets by their resource UUID
+- **`#field` syntax** — read specific fields (username, password, URI, TOTP, custom fields)
+- **Bulk loading** with `passboltBulk()` to load all passwords from a folder
+- **Custom fields object** with `passboltCustomFieldsObj()` for `@setValuesBulk`
+- **Self-hosted support** — API URL is read from your account kit
+- **Multiple instances** — connect to different Passbolt instances
+
+## Installation and setup
+
+In a JS/TS project, you may install the `@varlock/passbolt-plugin` package as a normal dependency.
+Otherwise you can just load it directly from your `.env.schema` file, as long as you add a version specifier.
+See the [plugins guide](/guides/plugins/#installation) for more instructions on installing plugins.
+
+```env-spec title=".env.schema"
+# 1. Load the plugin
+# @plugin(@varlock/passbolt-plugin)
+#
+# 2. Initialize the plugin - see below for auth setup
+# @initPassbolt(accountKit=$PB_ACCOUNT_KIT, passphrase=$PB_PASSPHRASE)
+# ---
+
+# 3. Add config items for your credentials
+# @type=passboltAccountKit @sensitive
+PB_ACCOUNT_KIT=
+# @sensitive
+PB_PASSPHRASE=
+```
+
+### Authentication setup
+
+The plugin authenticates using your Passbolt **account kit** (a base64-encoded bundle containing your server URL, keys, and user info) and your **passphrase** (used to decrypt your private key).
+
+<Steps>
+1. **Download your account kit** from Passbolt:
+    - Open your Passbolt instance
+    - Navigate to **Manage account** → **Desktop app setup**
+    - Click **Download your account kit** and copy the base64 contents
+
+2. **Wire up your credentials in the schema**:
+
+    ```env-spec title=".env.schema"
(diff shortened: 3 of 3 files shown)


theoephraim:
Is using the openapi generated thing really necessary? I suspect the number of api endpoints actually needed for the integration is minimal and easy enough to do more manually. Trying to keep this plugins fairly thin and dependencies minimal is important.

Note that we have been using [ky](https://www.npmjs.com/package/ky) in a few plugins to help simplify fetch interactions, but its very small and very widely used.

PaddeK:
It just was a fast way to get it working. I am working already on replacing everything not needed for the plugin. After that i will take care of the copilot findings. Thanks for having a look!

PaddeK:
I should have fixed everything mentioned by @theoephraim and copilot.

packages/plugins/passbolt/README.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+## Bulk Loading Secrets
+
+Use `passboltFolder()` with `@setValuesBulk` to load all secrets from a folder at once:
+
+```env-spec title=".env.schema"
+# @plugin(@varlock/passbolt-plugin)
+# @initPassbolt(accountKit=$PB_ACCOUNT_KIT, passphrase=$PB_PASSPHRASE)
+# @setValuesBulk(passboltFolder("CI\/CD/DEV"))
```
theoephraim: What's up with this backslash? Is that how it appears in the passbolt UI?

packages/plugins/passbolt/README.md
```diff
@@ -0,0 +1,219 @@
...
+- `passbolt(resourceId)` - Fetch by resource UUID from default instance
+- `passbolt(instanceId, resourceId)` - Fetch from a specific instance
+
+**Resource ID Format:**
+
+- Must be a valid UUID v4: `"12345678-1234-4567-abcd-123456789abc"`
+
+---
+
+#### `passboltFolder()`
```
theoephraim: I'm not familiar with passbold, but would it make sense to combine these two into a single `passboltBulk` method that can take a `folderPath` or a `resourceId` or potentially other filters? Not necessarily saying it has to be this way - whatever is sort of normal/idiomatic for passbolt - but it may provide more flexibility to search for the specific fields to fetch.

theoephraim:
This is awesome. Thanks so much :) Left a couple small questions but otherwise seems ready to go!

packages/plugins/passbolt/README.md
PaddeK: I have a folder `CI/CD` in my passbolt instance. Because i wanted to enable users to define a deeper nested folder structure as a argument i decided to use filepath like declrataion. This requires to escape the slash in the `CI/CD`. Just wanted to use something that is common as delimiter and escape char.

packages/plugins/passbolt/README.md
PaddeK: Thats a good idea to combine them. I could also include a `field` filter. This filter would only be useful for single resources (passbolt()) or for the bulk variant with folderPath though. I am not sure to ignore the `field` argument in the bulk variant with resourceId then or its better to have two bulk functions in that case to not confuse users.

PaddeK: To explain my line of thinking in more detail, i should briefly describe how Passbolt organizes information. There is a folder structure with references to parent folders and resources. A folder can contain multiple resources. Resources contains fields like username, url, password or custom fields. Custom fields is a way to create multiple key-value pairs inside a single resource.

packages/plugins/passbolt/README.md
theoephraim: Ahhh that makes sense. Although let’s use a different example then since the having a slash in the name is an unusual case.

theoephraim:
Ok I was looking at the passbolt docs, and I'm not 100% certain of everything but I think I understand a bit more about what's going on. Please forgive me if I'm way off...

It seems like a "resource" is an item that contains multiple fields (username/password/totp/url/custom fields?) while a "secret" is an individual field (just the username/pass/etc). Is it easy within the UI to get these "secret" ids?

The current implementation assumes you are always fetching the password field. That may be correct most of the time, but I think you need to allow the user to specify. For example you may have a username and password and need to fetch both. It's unclear to me how custom fields work from the api docs, but likely you want to allow referring to custom fields this way too.

```bash
ITEM_USERNAME=passbolt(resource-uuid#username)
ITEM_PASSWORD=passbolt(resource-uuid)    # <- defaults to password, similar to what you have now
ITEM_PASSWORD_EXPLICIT=passbolt(resource-uuid#password)
ITEM_CUSTOM_FIELD=passbolt(resource-uuid#custom-field-key) # not sure about this??
```

For bulk fetching, I guess you could fetch a bunch of items and only fetch their password fields, but I'm not sure that's quite right.

You could instead allow fetching all custom fields on a single item, something like:
```
# @setValuesBulk(passbolt(resource-uuid, customFieldsObj=true))
```
This would be useful say if you had a single item that had all your env vars as custom fields, stored under their correct names. Whether that should be a new resolver or just an option, not 100% sure...

You could of course also store a big dotenv style blob within a single "password" field (assuming it can handle long multi-line items).

Looks like there is also functionality to fetch items using categories (seems like folders?) or tags - but the fact the items are meant to store multiple fields makes bulk fetching this way make less sense.



Note that we have a few other plugins that have similar setups and we have used `id#key` as a way to refer to specific child keys within a larger group of stuff, especially when the key is optional.

PaddeK:
You are correct, the resource is a container for uri/username/password/totp/custom fields. Getting the resourceId (a UUID) is simple via the UI and described in the readme. 

I had a look at the AWS Secrets Manager plugin and think the optional key handling is a good fit for the `passbolt()` single item resolver. But i dont see much use of a bulk resolver that is fetching all usernames or uri fields etc.

> ```
> # @setValuesBulk(passbolt(resource-uuid, customFieldsObj=true))
> ```

That usecase is covered with the `passboltCustomFields()` resolver. Personaly i will only use the bulk resolver for folders because it has better handling in the UI. 

>You could of course also store a big dotenv style blob within a single "password" field (assuming it can handle long multi-line items).

Handling for that is not good in the UI and was the reason i looked into custom fields.

>Looks like there is also functionality to fetch items using categories (seems like folders?) or tags - but the fact the items are meant to store multiple fields makes bulk fetching this way make less sense.

Categories and Tags are kinda newish (and only available in Pro or Cloud version) and the API does not return this information for a resource.

I think i will keep the dedicated `passboltCustomFields()` resolver for clarity and implement an optional field argument (like the AWS plugin) for the `passbolt()` and `passboltFolder()` resolvers with default password. For accessing custom field keys i would suggest something like this:

```
CUSTOM_FIELD_1=passbolt(resource-uuid#custom) # returns value of CUSTOM_FIELD_1 key
CUSTOM_FIELD_2=passbolt(resource-uuid#custom.CUSTOM_FIELD_1) # will also return value of CUSTOM_FIELD_1
```
I dont see much use for the second example but it might be useful to someone. The dot as delimiter is to distinguish between lets say `username` of the resource and possibly a custom field key `username` .

I also thought it could be nice to generate the TOTP code, the API only gives access to TOTP settings like digits/hash algorithm/secret/period and the code is only shown in the UI.

PaddeK:
Just pushed discussed changes and updated the README accordingly. Let me know what you think.

theoephraim:
@PaddeK - made a few small adjustments, but about to merge.

- no more "#custom.xyz" needed `passbolt(ID#xyz)` assumes custom field if name is not known field
- no more "#custom" inferring from the item name. Doesn't really make sense IMO
- renamed bulk resolver, to leave room for more filters in the future - `passboltBulk(folderPath="TestFolder")`
- renamed custom fields resolver `passboltCustomFieldsObj(UUID)` just to make clearer
- reworked auth flow so it only logs in once, not on every fetch
- added a whole test suite!

Sidenote - a bit awkward IMO to fetch only the password field when doing the bulk fetch, but there isn't really another good option. I guess it could see if the item only has a single field filled and always use that? Try it out for a bit and we can see what feels right...

PaddeK:
> @PaddeK - made a few small adjustments, but about to merge.
> 
> * no more "#custom.xyz" needed `passbolt(ID#xyz)` assumes custom field if name is not known field
> * no more "#custom" inferring from the item name. Doesn't really make sense IMO
> * renamed bulk resolver, to leave room for more filters in the future - `passboltBulk(folderPath="TestFolder")`
> * renamed custom fields resolver `passboltCustomFieldsObj(UUID)` just to make clearer
> * reworked auth flow so it only logs in once, not on every fetch
> * added a whole test suite!
> 
> Sidenote - a bit awkward IMO to fetch only the password field when doing the bulk fetch, but there isn't really another good option. I guess it could see if the item only has a single field filled and always use that? Try it out for a bit and we can see what feels right...

First of all thanks for the changes! 

I also thought about the bulk fetching with just one field as parameter. I also don't see another good option and the custom field resolver is good alternative for users. Looking for the only set field might be a good option with fallback to the given field. That behaviour could be opt-in via a second parameter "autoinfer" or something like that.

Another though i had would require some changes to the plugin system. If the plugin had access at resolve time to the @type decorator of the item the plugin could in this situation infer the right field to return.

theoephraim:
re: exposing the type - thats an interesting idea. Shouldnt be too hard, we already expose the item key.

But at the end of the day, the bulk stuff is really just a convenience and only makes sense for some plugins where it's somewhat normal for the user to store env vars with their exact names... Lets see what you (and other users) say after using it for a bit

PR #501, opened 2026-03-28
feat: add Dashlane plugin

LucasPicoli:
Adds @varlock/dashlane-plugin wrapping the Dashlane CLI (dcli) to resolve secrets via dl:// references. Supports pre-authenticated dcli sessions and headless auth via service device keys, multiple instances, and in-session caching.

### DIFF
--- a/packages/plugins/dashlane/package.json
+++ b/packages/plugins/dashlane/package.json
@@ -19,7 +19,8 @@
   "scripts": {
     "dev": "tsup --watch",
     "build": "tsup",
-    "test": "vitest"
+    "test": "vitest",
+    "typecheck": "tsc --noEmit"
   },
   "keywords": [
     "varlock",
--- a/bun.lock
+++ b/bun.lock
@@ -305,7 +305,7 @@
     },
     "packages/plugins/passbolt": {
       "name": "@varlock/passbolt-plugin",
-      "version": "0.0.1",
+      "version": "0.0.0",
       "dependencies": {
         "openpgp": "6.3.0",
       },
--- a/packages/plugins/dashlane/tsconfig.json
+++ b/packages/plugins/dashlane/tsconfig.json
@@ -1,14 +1,5 @@
 {
-  "compilerOptions": {
-    "target": "es2021",
-    "module": "esnext",
-    "moduleResolution": "bundler",
-    "strict": true,
-    "resolveJsonModule": true,
-    "esModuleInterop": true,
-    "skipLibCheck": false,
-    "customConditions": ["ts-src"]
-  },
+  "extends": "@varlock/tsconfig/plugin.tsconfig.json",
   "include": ["**/*.ts"],
   "exclude": ["node_modules", "dist"]
 }


philmillman:
Thanks for this! Will review soon

theoephraim:
@LucasPicoli - would you be using this for local dev only? Doesn't seem like they have an sdk/api to pull from in production without using the CLI - which would mirror what we do for 1pass (use sdk with service account tokens for prod)

packages/plugins/dashlane/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@varlock/dashlane-plugin",
+  "description": "Varlock plugin to load secrets from Dashlane via dcli",
+  "version": "0.0.1",
```
theoephraim: leave it at 0.0.0 - our publishing automation will bump it.

packages/plugins/dashlane/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# @varlock/dashlane-plugin
```
theoephraim: remove this file. Instead use a changeset (run `bun run changeset:add` at the root), make it a patch, and include your summary there. The publishing automation takes care of creating the changeset from that file.

packages/plugins/dashlane/README.md
```diff
@@ -0,0 +1,277 @@
+# @varlock/dashlane-plugin
+
+This package is a [Varlock](https://varlock.dev) [plugin](https://varlock.dev/guides/plugins/) that enables loading secrets from [Dashlane](https://www.dashlane.com/) into your configuration. It wraps the [Dashlane CLI (`dcli`)](https://github.com/Dashlane/dashlane-cli) to resolve secrets via `dl://` references.
+
```
theoephraim, line 4: might want to add an extra note here to really call out that this requires `dcli` to be installed and in the user's PATH.

packages/plugins/dashlane/README.md
```diff
@@ -0,0 +1,277 @@
...
+#### `dashlaneDeviceKeys`
+
+Validates the `dls_*` format for Dashlane service device keys. Marked as sensitive.
+
+#### `dashlaneSecretRef`
+
+Validates the `dl://` URI format for Dashlane secret references.
+
+```env-spec title=".env.schema"
+@type(dashlaneDeviceKeys)
```
theoephraim:
```suggestion
# @type(dashlaneDeviceKeys)
```

packages/plugins/dashlane/README.md
```diff
@@ -0,0 +1,277 @@
...
+DASHLANE_SERVICE_DEVICE_KEYS=
+
+@type(dashlaneSecretRef)
```
theoephraim:
```suggestion
# @type(dashlaneSecretRef)
```

theoephraim: also this is probably unlikely to be used in most cases?

packages/plugins/dashlane/README.md
```diff
@@ -0,0 +1,277 @@
...
+MY_SECRET_REF=
+```
+
+---
+
+## How it works
+
+Under the hood, the plugin:
+
+1. Executes `dcli read <dl://reference>` as a subprocess for each secret
```
theoephraim, line 209: does `dcli` provide any bulk reading mechanism? Are there any rate limits to worry about?

packages/plugins/dashlane/README.md
```diff
@@ -0,0 +1,277 @@
...
+2. Delegates authentication to `dcli` -- either via an existing interactive session or service device keys passed as the `DASHLANE_SERVICE_DEVICE_KEYS` environment variable
+3. Caches resolved values in memory for the duration of a single resolution session (no secrets are persisted)
+
+### Sync behavior
+
+`dcli read` works against a **local vault cache** that auto-syncs with Dashlane's servers once per hour. This means:
+
+- Secrets changed less than 1 hour ago may not be reflected without a manual sync
+- Run `dcli sync` to force a fresh sync before resolving
```
theoephraim, line 218: should we internally trigger `dcli sync` at the beginning of the resolution? Probably too aggressive... but maybe there is a way to check when it has been synced last and trigger it depending on that, to make it more often than 1 hr? or could add a `triggerSync` param within `@initDashlane` too?

packages/plugins/dashlane/README.md
```diff
@@ -0,0 +1,277 @@
+# @varlock/dashlane-plugin
```
theoephraim, line 1: Please add the docs page to the website as well. Other plugin pages should provide a good template/example.

packages/plugins/dashlane/src/plugin.test.ts
```diff
@@ -0,0 +1,384 @@
+import {
```
theoephraim: thanks for adding tests. I hope to do a pass soon at providing some tooling to make it easier for plugins to include tests.

LucasPicoli:
@theoephraim it does work for CI/prod too, not just local dev. Dashlane supports headless auth via https://cli.dashlane.com/personal/devices which the plugin accepts through serviceDeviceKeys, same bootstrap pattern as the 1pass plugin. The main difference is that they don't have any SDK or API, so it's CLI-only for both paths, similar to the pass plugin

packages/plugins/dashlane/README.md
LucasPicoli: unfortunately it does not, but we don't have to worry about it since vaults are local. the limits exist only on the initial sync, as it seems

packages/plugins/dashlane/README.md
LucasPicoli: agreed, will drop it

packages/plugins/dashlane/README.md
LucasPicoli: I considered this but thought it was a bit too much overhead since the reads try and make sure everything is synced. But yeah, I think it's better if we do it on init, just to make sure everything is up to date.

LucasPicoli:
just a note - I also noted that the dcli will not lock itself after working on it. this should be fine on interactive environments, especially if we are resolving multiple credentials. I'll make sure to add a note that tells people to lock their vaults on the instructions. however, for headless I'll make sure that we lock the vault down at the end.

packages/plugins/dashlane/README.md
theoephraim: do keep in mind that when doing live-reload style front-end dev, it may be reloading the config a lot. So it still probably makes sense as an option?

theoephraim:
re: locking - might want to make that an option too. The default can be set based on whether it is headless or not. This might require adding a new hook or something, as I'm not sure there will be a natural place to add it. I had initially imagined lots of hooks available to plugins like this but there hasn't really been a need yet.

Also one important thing - make sure that any init / connect logic only runs when there is actually a first read triggered. You dont want it to happen on plugin init, because you may initialize the plugin in your .env.schema, but only use it in some cases, and you dont want for example a missing `dcli` to cause a problem if no dashlane secrets are being read.

theoephraim (approved):

PR #512, opened 2026-03-30
add copilot workflow and instructions

philmillman:
instructions need to be in the first 4k characters for code review 

image


philmillman:
@theoephraim feel free to modify instructions

PR #564, opened 2026-04-07
pass through USER/HOME to op cli

theoephraim:
passes through USER and HOME env vars to op cli calls within 1password plugin. This is to fix op plugin behaviour on machines where op is installed via brew in a multi-user setup.

### DIFF
--- a/.changeset/bold-trees-relate.md
+++ b/.changeset/bold-trees-relate.md
@@ -1,0 +1,5 @@
+---
+"@varlock/1password-plugin": patch
+---
+
+pass through USER and HOME to op cli calls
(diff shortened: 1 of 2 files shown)


philmillman (approved):

theoephraim:
user reports they still need `OP_BIOMETRIC_UNLOCK_ENABLED` although I'm unclear why it's different on their system. We can probably just always pass through `OP_BIOMETRIC_UNLOCK_ENABLED=true` since this code path is only called when using the cli specifically for biometric auth.

philmillman:
> user reports they still need `OP_BIOMETRIC_UNLOCK_ENABLED` although I'm unclear why it's different on their system. We can probably just always pass through `OP_BIOMETRIC_UNLOCK_ENABLED=true` since this code path is only called when using the cli specifically for biometric auth.

is it always going to be biometric though? won't it just default to whatever the system level auth is?

theoephraim:
the docs claim the setting

> Toggles the [1Password app integration](https://developer.1password.com/docs/cli/app-integration#optional-set-the-biometric-unlock-environment-variable) on or off.

so... 🤷

philmillman:
> the docs claim the setting
> 
> > Toggles the [1Password app integration](https://developer.1password.com/docs/cli/app-integration#optional-set-the-biometric-unlock-environment-variable) on or off.
> 
> so... 🤷

gotcha, naming things is hard

PR #592, opened 2026-04-12
fix: Make execSyncVarlock work in a shell-less environment

TeaSeaLancs:
execSyncVarlock's initial execution, trying to find varlock on PATH, fails in a hardened environment where there is no shell installed with an ENOENT error.

Given that the ENOENT error from execSync is unique to when the shell itself is not found (ie: A missing varlock binary triggers a different error), capturing the ENOENT error and allowing a fallthrough to the rest of the function mitigates against this.

### DIFF
--- a/.changeset/chatty-beans-refuse.md
+++ b/.changeset/chatty-beans-refuse.md
@@ -1,0 +1,5 @@
+---
+"varlock": patch
+---
+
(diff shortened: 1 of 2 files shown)


theoephraim:
Thanks for this! Will have it publishing shortly :)

PR #622, opened 2026-04-16
feat(cloudflare): add SvelteKit + Cloudflare Vite plugin

theoephraim:
## Summary

- Adds `varlockSvelteKitCloudflarePlugin` exported from `@varlock/cloudflare-integration/sveltekit` for SvelteKit projects deployed to Cloudflare Workers via `@sveltejs/adapter-cloudflare`.
- The existing `varlockCloudflareVitePlugin` wraps `@cloudflare/vite-plugin`, which doesn't currently play nicely with SvelteKit ([cloudflare/workers-sdk#8922](https://github.com/cloudflare/workers-sdk/issues/8922)). The new variant skips that wrapper while still injecting a `cloudflare:workers` env loader into the SvelteKit SSR entry (`@sveltejs/kit/.../runtime/server/index.js`, which adapter-cloudflare's generated `_worker.js` imports).
- The injection is guarded by `navigator.userAgent === 'Cloudflare-Workers'` and uses a dynamic import so SvelteKit's Node-side postbuild steps (prerender, fallback) don't fail resolving `cloudflare:workers`.
- Extracts the shared env-loader body (`__VARLOCK_ENV` / chunked bindings → `globalThis.__varlockLoadedEnv`) into `shared-ssr-entry-code.ts` and reuses it from both plugin variants. The existing Workers-only plugin keeps its synchronous static import — no runtime behavior change there.

## Why two variants instead of a flag

The two paths have effectively no overlap at runtime (different entry module, different dev story, different Vite plugin composition). A separate subpath export keeps the docs, errors, and types clean.

## Test plan

- [x] `bun run build` in `packages/integrations/cloudflare`
- [x] `npm run build` on the `sveltekit-on-cloudflare` example — SSR entry contains the guarded loader, static `import 'cloudflare:workers'` is preserved as external
- [x] `bun run lint:fix` clean
- [ ] End-to-end deploy of the SvelteKit example via `varlock-wrangler deploy` and verify env resolves at runtime in a real Worker
- [ ] Sanity-check existing Cloudflare Workers flow still works (snippet output is byte-identical to before this PR)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


philmillman (approved):
lgtm, maybe concern generic sveltekit instructions as discussed too

PR #624, opened 2026-04-16
feat(1password): bridge `op` CLI for local devcontainer

theoephraim:
## Summary

Inside a devcontainer the `op` CLI can't reach the host's 1Password desktop app for biometric auth, forcing users to fall back to service account tokens. This PR adds a `varlock-op-bridge` binary shipped with `@varlock/1password-plugin` that runs on the host and proxies `op` invocations over TCP (or a Unix socket). The plugin detects `VARLOCK_OP_BRIDGE_SOCKET` and routes every `op` call through the bridge transparently — so `op` doesn't need to be installed in the container and host biometric auth keeps working.

- New `varlock-op-bridge` CLI with `serve` + idempotent `ensure` (suitable for devcontainer `initializeCommand`)
- Token-based auth: 32-byte random token rotated per `ensure`, 0600 on host, bind-mounted read-only into the container
- TCP and Unix-socket transports; TCP is the recommended default (Docker Desktop on macOS has known issues bind-mounting Unix sockets)
- Env-blocklist server-side so the container can't clobber the host's `HOME`/`USER`/`PATH` when running `op`
- argv[0] renaming so 1Password's auth prompt shows `varlock-op-bridge` instead of `node`

## Architecture

```
┌───────────────── container ──────────────────┐          ┌──────────── host ────────────┐
│  varlock load                                │          │  varlock-op-bridge serve     │
│    └─ @varlock/1password-plugin              │  TCP     │    └─ spawn op               │
│         └─ invokeOpViaBridge(argv, env) ─────┼──────────┼─▶ run, stream result back    │
│             attaches token from env/file     │          │                              │
└──────────────────────────────────────────────┘          │  1Password desktop app       │
                                                          └──────────────────────────────┘
```

Single chokepoint in the plugin ([cli-helper.ts](packages/plugins/1password/src/cli-helper.ts)): if `VARLOCK_OP_BRIDGE_SOCKET` is set, route through the bridge; else spawn `op` locally. No behavior change for existing users.

## Setup (user-facing)

> To be turned into proper docs later — keeping it here so the PR is self-contained.

### Prereqs on the host

- 1Password desktop app installed and unlocked
- CLI integration enabled (**Settings → Developer → Integrate with 1Password CLI**)
- Docker Desktop running

No host install of anything else is required — `varlock-op-bridge` ships with `@varlock/1password-plugin` and can be invoked via `npx`/`pnpm dlx`/`bunx`.

### Add to your `devcontainer.json`

```jsonc
{
  // Runs on the HOST before the container starts; spins up the bridge if not
  // already running and rotates the shared-secret token. Idempotent.
  "initializeCommand": "npx -y -p @varlock/1password-plugin varlock-op-bridge ensure --addr 0.0.0.0:7195",

  // Inside the container, point the plugin at the host bridge and the token.
  "containerEnv": {
    "VARLOCK_OP_BRIDGE_SOCKET": "host.docker.internal:7195",
    "VARLOCK_OP_BRIDGE_TOKEN_FILE": "/run/varlock-op-bridge.token"
  },

  // Bind-mount the token file (0600 on host) read-only into the container.
  "mounts": [{
    "source": "${localEnv:HOME}/.varlock-op-bridge.token",
    "target": "/run/varlock-op-bridge.token",
    "type": "bind",
    "readonly": true
  }],

  // Linux hosts need this so `host.docker.internal` resolves. Harmless on macOS/Windows.
  "runArgs": ["--add-host=host.docker.internal:host-gateway"]
}
```

The other package-manager variants of the `initializeCommand`:

```jsonc
// pnpm
"initializeCommand": "pnpm dlx -p @varlock/1password-plugin varlock-op-bridge ensure --addr 0.0.0.0:7195"
// bun
"initializeCommand": "bunx --package @varlock/1password-plugin varlock-op-bridge ensure --addr 0.0.0.0:7195"
```

### Why TCP on `0.0.0.0:7195`?

- Docker Desktop on macOS has [chronic issues](https://github.com/docker/for-mac/issues?q=is%3Aissue+bind+source+path+socket) bind-mounting Unix sockets, so TCP over `host.docker.internal` is the portable default.
- Must bind to `0.0.0.0` (not `127.0.0.1`) because container connections arrive from Docker's VM gateway IP (e.g. `192.168.65.254`), not loopback.
- macOS's default firewall blocks inbound from the physical network, so `0.0.0.0` doesn't widely expose the port. Still, the token auth is the real gate.
- If you prefer Unix sockets (Linux hosts, GitHub Codespaces, etc.), pass `--addr /tmp/varlock-op-bridge.sock` and use the same path for `VARLOCK_OP_BRIDGE_SOCKET` + the `mounts` source.

### Confirming it's working

Inside the container:

```bash
# Should print resolved secrets, biometric prompt happens on host once per session
pnpm exec varlock load --format json
```

First call per 1Password session pops a biometric prompt **on the host** — it says "varlock-op-bridge wants to access 1Password" (not "node"). Approving once covers all subsequent calls until 1Password's session expires.

### Troubleshooting

- **`connect ECONNREFUSED 192.168.65.254:7195`** — bridge is bound to 127.0.0.1 instead of 0.0.0.0. Use `--addr 0.0.0.0:7195` in `initializeCommand`.
- **`invalid mount config for type "bind": bind source path does not exist: /socket_mnt/...`** — you're on macOS using Unix-socket mode; switch to TCP.
- **`cannot create directory "/home/<user>/.config/op"`** — container is leaking its `HOME` to the host. Fixed server-side via env-blocklist; pull latest.
- **`op bridge error: ENOENT: \`op\` not found on host`** — install the `op` CLI on the host.
- **`unauthorized: invalid or missing bridge token`** — the `mounts` entry for the token file is missing, or the `VARLOCK_OP_BRIDGE_TOKEN_FILE` env var path is wrong.
- **Bridge not coming up** — check `~/.varlock-op-bridge.log` on the host.

## Security notes

- Host-local processes can read `~/.varlock-op-bridge.token` if they have the user's UID (0600 file perms). Same threat model as SSH keys.
- Sibling containers on the same host cannot auth without the token-file mount.
- Network attackers are blocked by macOS's default firewall on the bridge port.
- The bridge is a single shared `op` session for any caller with the token — it doesn't preserve 1Password's per-caller prompts. See discussion in [src/bridge/server.ts](packages/plugins/1password/src/bridge/server.ts).

## Test plan

- [x] Unit-level smoke test: server + client round-trip with a fake `op` on PATH (success, non-zero exit → `ExecError`, env-overlay, ENOENT)
- [x] Token auth: rejects missing/wrong tokens, accepts correct token
- [x] `ensure` idempotent: fresh start, reuse, stale socket cleanup, token rotation restarts old bridge
- [x] TCP mode bound to 0.0.0.0 reachable via `host.docker.internal` inside devcontainer (verified end-to-end in a throwaway repo)
- [x] 1Password prompt shows "varlock-op-bridge" after argv[0] trick
- [x] Local `bun run typecheck`, `bun run build`, and `bun run lint:fix` all clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/1password-bridge.md
+++ b/.changeset/1password-bridge.md
@@ -1,0 +1,11 @@
+---
+"@varlock/1password-plugin": minor-isolated
+---
+
+Add bridge mode for devcontainer / remote environments.
+
+The 1Password CLI normally can't reach the host's desktop app from inside a devcontainer, forcing users to fall back to service account tokens. This adds a `varlock-op-bridge` binary (shipped with the plugin) that runs on the host and proxies `op` invocations over TCP or Unix socket. The plugin detects `VARLOCK_OP_BRIDGE_SOCKET` and routes through the bridge transparently — so `op` doesn't even need to be installed inside the container, and host biometric auth still works.
+
+- New `varlock-op-bridge` CLI with `serve` and `ensure` subcommands (idempotent, suitable for devcontainer `initializeCommand`)
+- Token-based auth (32-byte random token rotated per `ensure`, 0600 on host, bind-mounted read-only into the container)
+- Supports TCP (`--addr host:port`) and Unix socket (`--addr /path.sock`); TCP recommended for Docker Desktop on macOS
--- a/packages/plugins/1password/README.md
+++ b/packages/plugins/1password/README.md
@@ -90,6 +90,8 @@ When enabled, if the service account token is empty, the plugin will use the des
 Keep in mind that this method connects as _YOU_ who likely has more access than a tightly scoped service account. Consider only enabling this for non-production secrets.
 :::
 
+> **Using this from a devcontainer?** The plugin ships a `varlock-op-bridge` helper that proxies `op` calls to your host's desktop app, so biometric unlock works from inside the container. See [Using desktop app auth from a devcontainer](https://varlock.dev/plugins/1password/#using-desktop-app-auth-from-a-devcontainer).
+
 ### Connect server setup (self-hosted)
 
 If you are running a self-hosted [1Password Connect server](https://developer.1password.com/docs/connect/), you can authenticate using a Connect server URL and token:
--- a/packages/plugins/1password/package.json
+++ b/packages/plugins/1password/package.json
@@ -15,6 +15,9 @@
   "exports": {
     "./plugin": "./dist/plugin.cjs"
   },
+  "bin": {
+    "varlock-op-bridge": "./dist/bridge-cli.cjs"
+  },
   "files": ["dist"],
   "scripts": {
     "dev": "bun run copy-wasm && tsup --watch",
--- a/packages/plugins/1password/src/bridge/addr.ts
+++ b/packages/plugins/1password/src/bridge/addr.ts
@@ -1,0 +1,46 @@
+export type BridgeAddr = | { kind: 'unix'; path: string }
+  | { kind: 'tcp'; host: string; port: number };
+
+/**
+ * Parses a bridge address string. Supports:
+ *   - Unix socket path:  "/tmp/varlock-op-bridge.sock", "~/foo.sock"
+ *   - TCP address:       "host.docker.internal:4455", "127.0.0.1:4455", ":4455"
+ *
+ * TCP detection rule: contains a `:` AND the portion after the last `:`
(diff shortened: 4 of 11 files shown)


theoephraim:
closing for now - but will likely come in handy!

PR #625, opened 2026-04-16
feat(azure-key-vault): add JSON key extraction to azureSecret()

bram209:


packages/plugins/azure-key-vault/src/plugin.ts
```diff
@@ -390,6 +405,29 @@ class AzurePluginInstance {
       });
     }
   }
+
+  async getSecret(secretRef: string, jsonKey?: string): Promise<string> {
+    const rawValue = await this.fetchSecretValue(secretRef);
+
+    if (jsonKey) {
```
bram209, line 412: I followed the pattern from the other plugins, but this logic is not really key-vault specific. Is there a shared utilities package for plugins, or would it make more sense to lift JSON key extraction into the core so each plugin doesnt have to reimplement it?

packages/plugins/azure-key-vault/src/plugin.ts
```diff
@@ -39,6 +39,7 @@ class AzurePluginInstance {
   private clientId?: string;
   private clientSecret?: string;
   private cachedToken?: CachedToken;
+  private secretCache = new Map<string, Promise<string>>();
```
bram209, line 42: same comment as below, if JSON key extraction would be lifted out of plugins, a simple per-run cache could live in the same place / not implemented within each plugin

theoephraim: we are soon rolling out an actual cache primitive, that caches across loads. Although it will be opt-in for users, so a single run cache might still make sense for some plugins. Could definitely think about making this built in and providing some utilities.

theoephraim:
Thanks for this - will have a think about making it a bit more built-in. Might merge this in the meantime but will see

bram209:
> Thanks for this - will have a think about making it a bit more built-in. Might merge this in the meantime but will see

Would appreciate a merge 🙏 , at work we want to start using varlock, but mainly have JSON secrets in our key vaults. But up to you of course!

theoephraim:
K will get this in and answer the larger questions later!

theoephraim (approved):

PR #637, opened 2026-04-21
chore(varlock-wrangler): use exit-code if set from process to avoid silent fails

labithiotis:
The varlock-wrangler cli was silently exiting with 0, even though wrangler had errors. 

Currently, the wrangler CLI would throw an error like below, but because the wrapper code returns exit 0, the CI/CD actions would pass.
```
✘ [ERROR] A request to the Cloudflare API failed.
Authentication error [code: 10000]
```

This PR sets the exit code from the process if present (not null), otherwise it would fall back to 0.

### DIFF
--- a/.changeset/slow-crabs-vanish.md
+++ b/.changeset/slow-crabs-vanish.md
@@ -1,0 +1,5 @@
+---
+"@varlock/cloudflare-integration": patch
+---
+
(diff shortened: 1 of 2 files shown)


philmillman (approved):
Thank you for this @labithiotis !

theoephraim (approved):

PR #653, opened 2026-04-23
fix(1password): forward proxy env vars in batch read subprocess

benevolent-tenacious-t:
## Problem

The `executeReadBatch` function in the 1Password plugin spawns `op run` with a minimal, hand-picked environment (just `PATH`, `USER`, `HOME`, `XDG_CONFIG_HOME`, and `OP_BIOMETRIC_UNLOCK_ENABLED`). This intentionally avoids leaking secrets into the subprocess — but it also strips `http_proxy` / `https_proxy` / `ALL_PROXY` / `NO_PROXY` and their case variants.

In any proxied environment, this causes `op` to bypass the proxy and attempt a direct connection, which fails:

```
error resolving value: Error: 1Password CLI error - error initializing client:
  Get "https://my-team.1password.com/api/v2/account/keysets":
  dial tcp: lookup my-team.1password.com: no such host
```

Environments affected:

- **Claude Code** — routes all sandboxed traffic through a localhost HTTP/SOCKS proxy; without proxy vars, DNS resolution itself fails
- **Corporate HTTP proxies** — common in enterprise environments
- **Docker / container networks** — when proxy vars are set for outbound access
- **CI runners behind proxies**

Note: the single-command path (`execOpCliCommand`) already preserves proxy vars because it uses `{ OP_SERVICE_ACCOUNT_TOKEN: _, ...cleanEnv } = process.env`. This PR brings the batch read path into parity.

## Fix

Add a `pickProxyEnv()` helper that selectively forwards the 8 standard proxy environment variables (both lowercase and uppercase forms) when present. Spread it into the `env` object alongside the existing passthrough vars.

## Testing

Verified in Claude Code's sandbox environment:
- **Before**: `varlock run --path varlock.schema -- echo test` fails with DNS error on `my-subdomain.1password.com`
- **After**: resolves the 1Password secret successfully, `op` connects through the proxy
- **Without proxy**: no behavior change — `pickProxyEnv()` returns `{}` when no proxy vars are set

### DIFF
--- a/.bumpy/fix-1password-proxy-env-653.md
+++ b/.bumpy/fix-1password-proxy-env-653.md
@@ -1,0 +1,5 @@
+---
+"@varlock/1password-plugin": patch
+---
+
+Forward proxy environment variables (`http_proxy`, `https_proxy`, `ALL_PROXY`, `NO_PROXY` and case variants) to the `op` subprocess in the batch read path. Fixes secret resolution failures in proxied environments (corporate proxies, Claude Code sandbox, Docker, CI runners behind proxies).
--- a/packages/plugins/1password/src/cli-helper.ts
+++ b/packages/plugins/1password/src/cli-helper.ts
@@ -12,10 +12,14 @@ const OP_CLI_CACHE: Record<string, any> = {};
 /** Proxy env vars that must be forwarded so `op` can reach 1Password through HTTP/SOCKS proxies
  * (e.g. corporate proxies, container networks, sandboxed dev tools like Claude Code). */
 const PROXY_ENV_KEYS = [
-  'http_proxy', 'HTTP_PROXY',
-  'https_proxy', 'HTTPS_PROXY',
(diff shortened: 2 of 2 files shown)


theoephraim:
This makes total sense. Might even want to migrate this into varlock core as some kind of helper (or at least a list) because I suspect we'll encounter this again with other plugins/situations.

Will get it figured out today.

Thanks so much!

benevolent-tenacious-t:
@theoephraim, I fixed the failing tests. Feel free to close this in favor of a varlock core if the helper make more sense.

benevolent-tenacious-t:
checking in to see if this has been included in a recent release in varlock core or if I still need a patch? @theoephraim

theoephraim:
Not yet! sorry for the delay. Just finishing up a huge release of built-in encryption utilities. Will get this and a few other small ones out shortly after!

theoephraim (approved):

PR #655, opened 2026-04-24
Feature: add folding region support for VSCode extension

madcampos:
Add support for folding regions using the `#region` and `#endregion` markers.

This is based on the [API docs for language extensions](https://code.visualstudio.com/api/language-extensions/language-configuration-guide#folding).

Reference for the feature in VSCode: https://code.visualstudio.com/docs/editing/codebasics#_folding

### DIFF
--- a/packages/vscode-plugin/language/language-configuration.json
+++ b/packages/vscode-plugin/language/language-configuration.json
@@ -31,5 +31,11 @@
     ["`", "`"],
     ["(", ")"],
     ["{", "}"]
-  ]
+  ],
+  "folding": {
+    "markers": {
+      "start": "^\\s*#?region\\b",
+      "end": "^\\s*#?endregion\\b"
+    }
+  },
 }


theoephraim:
so the user would actually have to add `# region` to their env files?

I think I'd rather set up a more concrete notion of sections and make those foldable. Please weigh in on [this discussion](https://github.com/dmno-dev/varlock/discussions/149). Once we have that figured out, we'll mark those as foldable.

madcampos:
> so the user would actually have to add `# region` to their env files?

@theoephraim that is correct. The user would have to set a `#region` and _only_ that provides folding regions on VSCode and other editors that support code folding in this way.

See the screenshot for reference:

image

---

> I think I'd rather set up a more concrete notion of sections and make those foldable. Please weigh in on [this discussion](https://github.com/dmno-dev/varlock/discussions/149). Once we have that figured out, we'll mark those as foldable.

From scanning the discussion, I think the two things are not quite aligned.
The discussion seems to me is more semantic, like html headers hierarchy, or nested json objects.

The `#region` folding is just to collapse code on the editor, like you would a code block between brackets. It has no semantics associated and would be removed when generating types and such.

theoephraim:
gotcha - I see that this explicit `// #region` type comments are supported for some other languages, so it makes sense. I figure most of the time you'd ideally be dealing with a "section" of some kind so figuring that out will help, but I dont see why we can't support this too.

madcampos:
Closing in favour of [this discussion](https://github.com/dmno-dev/varlock/discussions/149).
I've re-read it and I think we can advance both ideas together

PR #656, opened 2026-04-24
feat: encrypt injected env blob for secure serverless deployments

theoephraim:
## Summary

Adds support for encrypting the resolved env blob that varlock injects into server-side build output. This protects secrets from leaking via sourcemaps or build artifact inspection on serverless platforms (Vercel, Netlify, Cloudflare Workers, etc.).

### Encryption across all integrations
When `_VARLOCK_ENV_KEY` is set (or `@encryptInjectedEnv` decorator is enabled), the env blob is encrypted with AES-256-GCM before injection. At runtime, server-only init code decrypts it before `initVarlockEnv()` runs.

- **Vite / Astro / SvelteKit / TanStack Start** — encrypted blob set on `globalThis.__varlockEncryptedEnv`, decrypted in virtual init module
- **Next.js** — encrypted payload in `process.env.__VARLOCK_ENV`, decrypted by init-server/init-edge bundles
- **Cloudflare Workers** — encrypted blob uploaded as CF secret, decrypted in worker init
- **Expo** — no encryption needed (same-process)
- **auto-load** — encrypts `process.env.__VARLOCK_ENV` in-place, sets parsed object on globalThis

### Architecture
- Decryption removed from `env.ts` entirely — no `node:crypto` dependency in client-bundleable code
- New `globalThis.__varlockEncryptedEnv` separates encrypted blobs from parsed objects on `__varlockLoadedEnv`
- `crypto.ts` uses top-level `import` instead of broken `require` in ESM output

### New root decorators
- **`@encryptInjectedEnv`** — declares encryption intent, supports `forEnv()`. Auto-generates key for dev servers and Cloudflare deploys, errors on production builds without key
- **`@disableProcessEnvInjection`** — prevents injecting individual vars into `process.env`, forcing ENV proxy usage. Combined with encryption = zero plaintext secrets in `process.env`

### CLI changes
- `varlock generate-key` — fixed for ESM builds
- `varlock run --inject <all|vars|blob>` — controls what gets injected into child process env (replaces deprecated `--no-inject-graph`)

### Docs
- New "Encrypted deployments" guide with platform-specific setup (Vercel, Netlify, Cloudflare)
- Root decorator reference for both new decorators
- Integration pages updated with concise notes linking to guide

## Test plan
- [x] `varlock generate-key` works
- [x] Auto-load path encrypts/decrypts correctly
- [x] Auto-generates key when decorator set but no key provided
- [ ] Vite SSR build with `@encryptInjectedEnv` produces encrypted output
- [ ] Next.js build with `_VARLOCK_ENV_KEY` produces encrypted runtime files
- [ ] Cloudflare `varlock-wrangler deploy` auto-generates key
- [ ] `@disableProcessEnvInjection` prevents individual var injection
- [ ] `varlock run --inject blob` only injects blob
- [ ] Framework tests pass


framework-tests/frameworks/vite/vite-shared.ts
```diff
@@ -256,6 +256,38 @@ export function defineViteTests(
       });
     });
 
+    // ---- Encrypted env blob ----
+
+    describe('encrypted env blob', () => {
+      viteEnv.describeScenario('SSR build with _VARLOCK_ENV_KEY encrypts the blob', {
+        command: 'vite build --ssr src/ssr-entry.ts',
+        env: { _VARLOCK_ENV_KEY: '846a4cbdf4fefeff0da38d8f3766ffe50d8db12f8ce32849bb1e1a60ecb4ba0d' },
```
philmillman: should we make this dynamic?

packages/varlock-website/src/content/docs/integrations/nextjs.mdx
```diff
@@ -325,6 +325,51 @@ varlock run -- node .next/standalone/server.js
...
+    If you want local builds to also encrypt the blob (e.g., to test the flow), add the key to `.env.local`:
+
+    ```env title=".env.local"
+    _VARLOCK_ENV_KEY=your-64-char-hex-key-here
+    ```
+
+</Steps>
+
+:::tip
+You can define `_VARLOCK_ENV_KEY` in your `.env.schema` to enable validation (e.g., required in production). It will automatically be excluded from the injected blob and from type generation — it's infrastructure, not application config.
```
philmillman:
```suggestion
You can define `_VARLOCK_ENV_KEY` in your `.env.schema` to enable validation (e.g., required in production). It will automatically be excluded from the injected blob and from type generation.
```

packages/varlock-website/src/content/docs/integrations/nextjs.mdx
```diff
@@ -325,6 +325,51 @@ varlock run -- node .next/standalone/server.js
...
+
+```env-spec title=".env.schema"
+# @sensitive
+_VARLOCK_ENV_KEY=
+```
+:::
+
+:::note
+The encryption key is **never baked into the build**. It must always come from the runtime environment (e.g., a Vercel environment variable). The encrypted blob is useless without it.
```
philmillman:
```suggestion
The encryption key is **never baked into the build**. It must always come from the runtime environment (e.g., a Vercel environment variable). It is *required* for decryption.
```

packages/varlock-website/src/content/docs/integrations/vite.mdx
```diff
@@ -241,6 +241,57 @@ All non-sensitive items are bundled at build time via `ENV`, while `import.meta.
 
 
 
+---
+
+## Encrypting the env blob ||encrypting-the-env-blob||
+
+When using `ssrInjectMode: 'resolved-env'`, varlock injects the fully resolved env data into your SSR build output. By default, this blob is plaintext JSON — meaning anyone with access to the build artifact can read your secrets.
```
philmillman:
```suggestion
When using `ssrInjectMode: 'resolved-env'`, Varlock injects the fully resolved env data into your SSR build output. By default, this blob is plaintext JSON — meaning anyone with access to the build artifact can read your secrets.
```

packages/varlock-website/src/content/docs/integrations/vite.mdx
```diff
@@ -241,6 +241,57 @@ All non-sensitive items are bundled at build time via `ENV`, while `import.meta.
...
+You can define `_VARLOCK_ENV_KEY` in your `.env.schema` to enable validation. It will automatically be excluded from the injected blob and from type generation.
+
+```env-spec title=".env.schema"
+# @sensitive
+_VARLOCK_ENV_KEY=
+```
+:::
+
+:::note
+The encryption key is **never baked into the build**. It must always come from the runtime environment. The encrypted blob is useless without it.
```
philmillman:
```suggestion
The encryption key is **never baked into the build**. It must always come from the runtime environment. It is *required* for decryption.
```

philmillman (approved):
added a few nits. One thing we may want to consider for the future is whether we allow an arbitrary key name and then do the logic based on a type instead? Also does it make sense to allow them to generate/BYOK?

PR #567, opened 2026-04-08
varlock() native local encryption

theoephraim:
## Summary

Adds built-in device-local encryption to varlock via a new `varlock()` resolver function and native platform binaries. Encrypted secrets can be stored directly in `.env` files and are automatically decrypted at load time — no external secret manager required.

### Encryption backends

The best available backend is selected automatically at runtime:

| Platform | Backend | Key Storage | Biometric |
|----------|---------|-------------|-----------|
| macOS | Secure Enclave (Swift binary) | Hardware Secure Enclave | Touch ID / Face ID |
| Windows | DPAPI + Windows Hello (Rust binary) | Windows credential store | Windows Hello (face/fingerprint/PIN) |
| Linux | Secret Service + optional TPM2 layer (Rust binary) | GNOME Keyring / KWallet, sealed with TPM2 when available | Opt-in via polkit → PAM (fingerprint/face/YubiKey/password) |
| All | File-based fallback (pure JS) | \`~/.varlock/\` directory | No |

All backends use the same ECIES wire format (P-256 ECDH + HKDF-SHA256 + AES-256-GCM), so encrypted payloads are portable across implementations.

On macOS, the Secure Enclave provides hardware-backed keys that cannot be extracted, with optional biometric gating via Touch ID. On Windows, DPAPI protects keys scoped to the current user with optional Windows Hello biometric verification. On Linux, private keys live in the user's Secret Service keyring (GNOME Keyring / KWallet) and are additionally sealed to the TPM2 chip when one is available — so the "just works" path covers mainstream desktop distros, with defense-in-depth on hardware that supports it. TPM2-alone and plaintext tiers handle headless and minimal environments. A daemon process manages biometric sessions and IPC on all native platforms.

### Linux biometric unlock (opt-in)

A new \`setup --linux-biometrics\` subcommand installs a polkit policy that gates decrypts behind a PAM auth prompt. PAM picks up whatever factors the user has configured — fingerprint via fprintd, face via Howdy, YubiKey via pam_u2f, or just the login password. The daemon's existing TTY-scoped session cache wraps the polkit check, mirroring the Windows Hello flow (prompt on first decrypt per terminal, cached for 5 minutes).

### New CLI commands

- **\`varlock encrypt\`** — Encrypt values interactively or in bulk (\`--file .env.local\` encrypts all \`@sensitive\` plaintext values in-place)
[...]

PR #657, opened 2026-04-25
Add apple keychain support

BitPhinix:
I vibe-coded this entire thing but it's rather simple and works well

### DIFF
--- a/.bumpy/apple-keychain-plugin-init.md
+++ b/.bumpy/apple-keychain-plugin-init.md
@@ -1,0 +1,5 @@
+---
+"@varlock/apple-keychain-plugin": patch
+---
+
+Add apple keychain plugin
--- a/bun.lock
+++ b/bun.lock
@@ -171,6 +171,20 @@
         "varlock": "workspace:^",
       },
     },
+    "packages/plugins/apple-keychain": {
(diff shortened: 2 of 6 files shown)


theoephraim:
First off - thanks for the PR! :)

You sir, are in for a real treat. We have native keychain support baked into varlock core about to be released, and it's WAY better than deferring to the security cli :)

Theres a native swift binary that will be packaged with the npm module. It adds biometric fingerprint unlock, pinned to the current terminal session. Use `ITEM=keychain(prompt)` and you get a native mac popup to select an item, or add a new one. Automatically sets an ACL rule to limit access to said binary. Gets replaced in the file to to `ITEM=keychain(some/id)` and from then on resolves it normally (gated behind fingerprint).

Also we have a new `varlock()` resolver which uses secure enclave (behind fingerprint unlock) as an encryption key, but stores the encrypted value in the file, so no keychain involved.

You can install with the preview package available in [this PR](https://github.com/dmno-dev/varlock/pull/567). Use `https://pkg.pr.new/varlock@567` instead of a version number for varlock in your package.json. Would love to have some more feedback before it's released!

BitPhinix:
Oh amazing, thanks for your hard work!

PR #695, opened 2026-05-07
prompt builder and CLI agent mode

philmillman:


packages/varlock/src/cli/commands/load.command.ts
```diff
@@ -119,15 +126,32 @@ export const commandFn: TypedGunshiCommandFn<typeof commandSpec> = async (ctx) =
...
+    if (agent) {
+      const redactedEnv: Record<string, unknown> = {};
+      const resolvedEnv = envGraph.getResolvedEnvObject();
+      for (const itemKey of envGraph.sortedConfigKeys) {
+        const item = envGraph.configSchema[itemKey];
+        const value = resolvedEnv[itemKey];
+        if (item.isSensitive && typeof value === 'string') {
+          redactedEnv[itemKey] = redactString(value);
+        } else if (item.isSensitive && value !== undefined) {
+          redactedEnv[itemKey] = '[REDACTED]';
```
theoephraim: still might be useful to have the masked value showing a few chars? Maybe `AB**** [redacted]`?

theoephraim (approved):
looks good - I fixed a bunch of icons

packages/varlock/src/cli/commands/load.command.ts
philmillman: hmm weird, it's showing me a few chars when I test locally

image

theoephraim: I misread - it only applies to non-string values. I dont think that would really ever be tripped until we do more with objects/arrays/etc...

theoephraim (approved):

PR #721, opened 2026-05-26
fix(cloudflare): skip wrangler restart when env is unchanged

attilakis:
## Problem

`@varlock/cloudflare-integration` restarts wrangler whenever a watched env file emits an fs event, even if the file contents haven't changed, provided more than 5s has passed since the previous restart. In practice this wipes worker state on every save during dev (open `.env`, hit save twice, second save kills the worker).

## Root cause

PR #631 added a debounce + cache-equality check so unchanged-env saves were no-ops. PR #635 then layered a `FORCE_RESTART_IDLE_MS = 5000` override that re-triggers a restart after the idle window even when `freshLoaded.json === cachedGraphJson`, reintroducing the very behaviour #631 was meant to suppress.

`fs.watch` already fires on every save regardless of content, so the cache-equality check alone is sufficient. The idle threshold has no remaining job, if the env didn't change, wrangler has nothing to reload.

## Fix

Restart only when the resolved env JSON actually changes. Drop `FORCE_RESTART_IDLE_MS`, `lastRestartAt`, and `idleSinceLastRestart`.

### DIFF
--- a/.bumpy/cf-skip-restart-when-env-unchanged.md
+++ b/.bumpy/cf-skip-restart-when-env-unchanged.md
@@ -1,0 +1,5 @@
+---
+'@varlock/cloudflare-integration': patch
+---
+
+cloudflare: skip wrangler restart when env file is saved with unchanged contents (drops the 5s idle threshold that re-triggered restarts on every save after a short pause)
--- a/packages/integrations/cloudflare/src/varlock-wrangler.ts
+++ b/packages/integrations/cloudflare/src/varlock-wrangler.ts
@@ -441,12 +441,8 @@ async function handleDev(args: Array<string>) {
 
   // watch env source files for changes and restart wrangler with fresh data
   const DEBOUNCE_MS = 300;
-  // force a restart if the last save event was more than this long after the previous restart,
-  // even when the resolved env is identical — handles repeated saves of an unchanged file
-  const FORCE_RESTART_IDLE_MS = 5000;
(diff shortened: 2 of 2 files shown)


theoephraim:
@attilakis - this was an intentional decision to mirror behaviour seen in other tools (vite). My intuition is that when a user is purposefully hitting save on a .env file, they want to see some feedback, even if they didn't change anything. That said, when the file is churning and it was not the user's doing, it probably feels wrong.

In your case, what is causing the env file updates?

I think a middle ground might be restarting if the _env file contents_ has changed, regardless of if the final vars / loaded json does. This means if you even update a comment it would restart, but if you just hit save without anything changing it would not. I just tested a fresh nextjs app and this is what it does.

Some reasoning -- lets say you have some env file, with `ITEM=op(op://some/1p/path)` and you change it to `ITEM=op(op://different/1p/path)`, but both items are holding the same value. You are hitting save, wondering why nothing is happening or if it's working at all... The reload gives you some feedback / confidence that things are actually being loaded.

theoephraim:
Ok I think I have something better now. I am checking the actual injected values rather than just file contents, but I am giving clear user feedback about what is happening. For example when no values changed:

```
[varlock-wrangler] change detected in 1 env source file; reloaded env, no changes found, skipping restart.
```

Will close this PR and get mine out shortly. Thanks for bringing it up.

Note that vite plugin behaviour (which includes using the cloudflare+vite setup) remains unchanged, mirroring vite's behaviour of restarting regardless.

PR #742, opened 2026-06-03
fix(vscode-plugin): highlight URL fragments in @docs() decorator

Shtian:
## Summary

When adding the item decorator @docs() with a URL that had a fragment, the fragment symbol itself `#` fell out of the syntax scope: 
image

## Fix
Removed `#` from the `unquoted-string-within-fn-call`. Since this rule is bounded by parenthesis it should hold no special meaning, unless I've missed a special case 👀 

Added some tests to verify the behaviour, as well as multiline fn-calls.

Some examples containing the deprecated docsUrl decorator was changed to docs

After implementing the fix:
image

Let me know if I missed or need to change something!

### DIFF
--- a/.bumpy/fix-vscode-url-fragment-highlighting.md
+++ b/.bumpy/fix-vscode-url-fragment-highlighting.md
@@ -1,5 +1,6 @@
 ---
 env-spec-language: patch
+"@env-spec/parser": patch
 ---
 
 Fixed URL fragments in @docs() decorator not being highlighted correctly.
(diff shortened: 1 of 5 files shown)


theoephraim:
Thanks for this! Will take a look tomorrow.

I just came across this too - https://github.com/johnsoncodehk/monogram
which might help keep the grammar and highlighting all in sync with a bit less work 🤞

theoephraim (approved):

PR #750, opened 2026-06-05
Static/dynamic config controls

theoephraim:
Adds first-class control over whether config items are static (eligible for build-time inlining) or dynamic (must stay runtime-resolved), decoupled from sensitivity. Previously the only way to keep a value out of build-time bundles was to mark it `@sensitive`.

## Schema

- `@dynamic` / `@static` item decorators (boolean-valued, mutually exclusive)
- `@defaultDynamic` root decorator: `true`, `false`, or `inferFromSensitive`
- Default is `inferFromSensitive`: dynamic follows sensitivity, so existing schemas behave exactly as before. The serialized blob only carries `isDynamic` when it diverges from `isSensitive`; consumers read `isDynamic ?? isSensitive`.

## Runtime + integrations

- All bundler inlining (vite, next webpack/turbopack, expo babel) now gates on `isDynamic` instead of `isSensitive`. Secrecy behavior (redaction, leak scanning, sourcemap scrubbing, wrangler secrets) stays on `isSensitive`. Expo warns at build time when any non-inlined (sensitive or dynamic) key is referenced from native code.
- Public+dynamic values load in the browser via `loadPublicDynamicEnv()` hydrating the same `ENV.KEY` proxy, served from a JSON endpoint (`getPublicDynamicEnv()`). Astro auto-injects `/__varlock/public-env` when public+dynamic keys exist (configurable via `publicDynamicEndpoint`); Next/SvelteKit add a one-liner route. Hydration is filtered to declared public+dynamic keys; every integration (including Next, via the webpack client runtime chunk / turbopack client components) injects the declared key list (names only) into client bundles.
- Next.js: accessing a public+dynamic key during server rendering marks the route dynamic (runtime access hook calling `headers()`), including nested component access. Works on both bundlers; a source-rewrite approach was dropped after breaking pages-router builds.
- Vite-based builds set `__VARLOCK_EXECUTION_PHASE=build`; accessing a public+dynamic key during build/prerender errors (downgradable via `_VARLOCK_DYNAMIC_BUILD_ACCESS_MODE=warn`). Sensitive values are deliberately not guarded: reading them server-side during SSG is legitimate, and leak scanning polices the output.

- Boot-time injection alternative to the endpoint: the browser runtime auto-hydrates from a `globalThis.__varlockPublicDynamicEnv` global if present, so a server/container entrypoint can extract the public+dynamic subset (`varlock load --filter '@dynamic,!@sensitive' --format json`) and inject it as a script tag into served HTML (recipe in the guide).

## Filtering

- New `@dynamic` filter selector (`!@dynamic` for static items) in `--filter` and codegen `filter=`.
- Decorator-selector filters (`@dynamic`/`@sensitive`/`@required`) now scope resolution and validation via metadata-first resolution: item decorator metadata resolves first (no value resolvers run), the filter matches exactly, and only selected items plus deps get values resolved. So `varlock load --filter='!@dynamic'` at build time skips runtime-only vars entirely, including their `@required` checks. A broken value that the filter evaluation itself depended on still fails the load.

## Validation

- varlock unit suite, plus framework tests: Next.js v16 (webpack + turbopack, incl. pages router), Astro v5, SvelteKit, Vite (dynamic non-inlining + HTML replacement), expo
- docs: new Static vs Dynamic guide (framework recipes centralized there in tabs, with boot-time-resolution and "when do you need this" framing), decorator/CLI/reserved-variable references, short per-integration sections


theoephraim:
@pullfrog review this PR

theoephraim:
@pullfrog - review again, we changed the model

PR #757, opened 2026-06-08
fix(parser): treat whitespace-only lines as blank lines

yinjs:
## Problem

A `.env` line containing only spaces/tabs dead-ended the parser. On varlock@1.4.0 this surfaced as a raw `Error: expected dataType to be set` crash; on newer versions it surfaces as a parse error. Standard dotenv parsers tolerate whitespace-only blank lines.

## Fix

Three grammar changes in `packages/env-spec-parser/grammar.peggy`:

- `_ "\n"` — whitespace-then-newline blank line (strict superset of the old `"\n"`)
- `__ !.` — whitespace-only final line at EOF with no trailing newline (uses `__` = one-or-more, so ≥1 char is consumed → no zero-width infinite loop)
- `_n = _ ("\n" / !.)` — line terminator now tolerates trailing horizontal whitespace, so a whitespace blank line directly after a comment/decorator line no longer dead-ends the parse

`src/grammar.js` is generated from `grammar.peggy` via `build:grammar` (peggy) and is gitignored, so it is not in the diff — CI regenerates it.

## Tests

7 new tests in `test/general.test.ts`: mid-file, trailing (no final newline), consecutive, leading, CRLF, post-comment, and post-decorator whitespace lines. All 7 fail on the unpatched grammar (load-bearing).

- `@env-spec/parser`: 217 passed
- `varlock`: 507 passed
- lint clean

Changeset: patch bump for `@env-spec/parser` + `varlock` (varlock bundles the parser).

### DIFF
--- a/.bumpy/parser-whitespace-blank-line.md
+++ b/.bumpy/parser-whitespace-blank-line.md
@@ -1,0 +1,6 @@
+---
+"@env-spec/parser": patch
+"varlock": patch
+---
+
+fix: treat whitespace-only lines as blank lines instead of throwing a parse error
--- a/packages/env-spec-parser/grammar.peggy
+++ b/packages/env-spec-parser/grammar.peggy
@@ -17,7 +17,8 @@ EnvSpecFile =
     ConfigItem
     / Divider
     / CommentBlock
-    / "\n" { return new ParsedEnvSpecBlankLine({ _location: location() }); }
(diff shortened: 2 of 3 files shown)


philmillman:
Thanks for this @yinjs. Will have a look today!

theoephraim:
seems reasonable - will get this merged today. Thanks!

theoephraim (approved):

PR #778, opened 2026-06-12
feat: per-item leak-detection opt-out via `@sensitive={preventLeaks=false}`

theoephraim:
## What

Lets a user opt a single secret out of runtime **leak detection** while keeping it sensitive everywhere else. Motivating case: an API endpoint that legitimately returns a secret to another system — today that trips varlock's response leak scanner.

```sh
# @sensitive={preventLeaks=false}
TOKEN_FORWARDED_TO_PARTNER=
```

The opt-out is **leak-detection only** — the value is still redacted in logs. Leak detection and log redaction previously shared one secrets map; this decouples them per-secret (the scanner skips opted-out entries, redaction still masks them).

## The `@sensitive={...}` options form

| option | type | default | meaning |
| --- | --- | --- | --- |
| `enabled` | boolean | `true` | whether the item is sensitive. May be a function for dynamic control (`enabled=forEnv(production)`); `{enabled=false}` ≡ `@sensitive=false`. |
| `preventLeaks` | boolean | `true` | when `false`, excluded from the response leak scanner (still redacted in logs). |

```sh
# sensitive only in production, and allowed to leave the system
# @sensitive={enabled=forEnv(production), preventLeaks=false}
PARTNER_TOKEN=
```

## How it flows

- **Schema** → `@sensitive={…}` parses to an object literal.
- **Graph** (`config-item.ts`) → reads `enabled` (→ sensitivity) and `preventLeaks` onto the item; rejects invalid uses (`@public={…}`, unknown option, non-boolean, array literal, bare `@sensitive(…)`).
- **Serialization** (`env-graph.ts`) → emits `preventLeaks: false` only when opted out.
- **Runtime** (`runtime/env.ts`) → `scanForLeaks` skips opted-out secrets; `redactSensitiveConfig` still masks them.

## New syntax: `{}` / `[]` literals

Carrying options on `@sensitive` required a value syntax. We chose **distinct brackets** — `{key=value}` objects and `[a, b, c]` arrays — keeping `()` reserved for function calls. Rationale (mainstream langs keep objects in `{}` not `()`; Lisp's all-`()` was walked back by Clojure; ML-family records already use `{k=v}`). Usable as decorator values and function-call args, including nested (`fn(retry={count=3})`).

Guidance on **when to use literals vs. not** (don't wrap a function's own named args) is documented in the env-spec reference and a grammar `DESIGN NOTE`.

## Deliberately deferred (documented)

- Standalone literals as **item values** (`FOO={…}`) — gated on object/array value handling (coercion/validation/redaction-of-objects), not a blocker but its own change.
- **Multi-line** literals.

## Tests

- Parser: object/array literal parse, nesting, round-trip.
- env-graph: opt-out keeps item sensitive + serializes the flag; static + dynamic (`forEnv`) `enabled`; rejection cases.
- Runtime: opted-out value passes the scanner, other secrets still throw, opted-out value still redacted.

All green: varlock + `@env-spec/parser` suites, typecheck, lint. Changeset included (`minor` for both packages).

### DIFF
--- a/.bumpy/sensitive-prevent-leaks-opt-out.md
+++ b/.bumpy/sensitive-prevent-leaks-opt-out.md
@@ -3,6 +3,6 @@
 varlock: minor
 ---
 
-Add per-item leak-detection opt-out via `@sensitive={preventLeaks=false}`. Secrets that legitimately leave the system (e.g. an API endpoint that returns a secret to another service) can be excluded from runtime leak detection while still being redacted in logs.
+Add per-item leak-detection opt-out via `@sensitive={preventLeaks=false}`. Secrets that legitimately leave the system (e.g. an API endpoint that returns a secret to another service) can be excluded from runtime leak detection while still being redacted in logs. The options form also accepts `enabled` to toggle sensitivity (including dynamically, e.g. `@sensitive={enabled=forEnv(production)}`).
 
 Adds standalone object (`{key=value}`) and array (`[a, b, c]`) literals to the env-spec grammar, usable as decorator values and function-call arguments (including nested). `()` remains reserved for function calls.
--- a/.bumpy/sensitive-prevent-leaks-opt-out.md
+++ b/.bumpy/sensitive-prevent-leaks-opt-out.md
@@ -1,0 +1,8 @@
+---
+"@env-spec/parser": minor
+varlock: minor
+---
+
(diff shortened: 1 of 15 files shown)


theoephraim:
Superseded by #783 — renamed the branch from the auto-generated `claude/epic-clarke-ddca4f` to a meaningful name (`sensitive-prevent-leaks-opt-out`) per AGENTS.md. Same commits.

PR #816, opened 2026-06-22
fix(vscode): highlight bare literal values and column-0 comments in literals

theoephraim:
## What

Fixes two gaps in the @env-spec TextMate grammar's handling of array/object literals.

**1. Bare item-value literals weren't highlighted.** `KEY=[a, b]` and `KEY={ k=v }` rendered as plain unquoted strings — the bare item-value rule never entered the bracket-literal rules (only `fn(...)` and decorator contexts did). These now highlight as proper array/object literals.

Since bare multi-line literals aren't valid @env-spec (they require a `fn()` wrapper), the bare context uses a single-line variant that closes at end of line, so an unclosed `KEY=[` no longer swallows the following entries.

**2. Column-0 comments inside multi-line `fn(...)` literals were mis-highlighted.** An unindented comment line inside a multi-line literal:

```
KEY=fn([
  ONE,
# this should be a comment
  TWO,
])
```

…had its `# ` eaten by the decorator-continuation marker, rendering the rest as a string. The literal rule tree handled a column-0 `#` the same way in every context, but it means opposite things: in a decorator (`# @dec(pick=[ … ])`) it's the comment-block prefix, while in an item value it's a real comment.

## How

Split the literal rule tree into decorator-context and item-context variants. The decorator variant keeps the structural-`#` prefix marker; the item variant drops it, so any `#` is a full comment. Item-context entry points route to the item/inline variants; decorator rules are unchanged.

## Tests

- **Assertion tests** (`test/grammar.test.txt`, run by `test:grammar`): bare inline arrays/objects, nesting, stray-bracket recovery, column-0/indented/trailing comments inside `fn(...)` literals, bare multi-line `fn(...)` calls (positional args, no bracket literal), and an unquoted decorator value terminated by `#` with no preceding space.
- **Snapshot fixtures** (`test/fixtures/*`): extended `.env.literals` and `.env.multiline` with the bare/column-0 cases and regenerated the snapshots. Also regenerated `.env.comprehensive.snap`, which was already stale (it expected a removed `source.bash.embedded` scope). These snapshots were **not run by any script or CI** before — added a `test:snap` script and wired it into `test:ci`, so they're now enforced.
- Manual highlighting-check file (`.env.example`) filled out with a dedicated array/object literal section.

All examples validated against the parser. Full suite (unit + grammar + snapshot) passes.


theoephraim:
hopefully a lot of this will be simplified with monogram!

PR #823, opened 2026-06-23
astro7 + cloudflare adapter compat

philmillman:
## Summary
- Fix ENV not initialized when using `@varlock/astro-integration` with the `@astrojs/cloudflare` adapter — varlock injects its init into the Cloudflare worker entry (`\0virtual:cloudflare/worker-entry`) with edge-runtime settings and binding-loader code, so `import { ENV } from 'varlock/env'` works in `astro dev` (workerd/miniflare) and in deployed Workers (via `varlock-wrangler deploy`). Requires `@varlock/cloudflare-integration` (optional peer).
- Share `CLOUDFLARE_SSR_ENTRY_CODE` from `@varlock/cloudflare-integration` so the Astro and Cloudflare integrations reuse the same string without duplication.
- **Suppress wrangler's redundant `.env` auto-loading** in the CF integrations (Astro adapter + standalone Vite plugin). `@cloudflare/vite-plugin` otherwise loads `.env` itself, printing `Using secrets defined in .env` and injecting raw values that shadow varlock's resolved env. We disable it via `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV` (respecting an explicit user override) and print a varlock notice instead.
- Silence Vite's "dynamic import cannot be analyzed" warning from varlock's runtime plugin loader by adding a `/* @vite-ignore */` magic comment alongside the existing `webpackIgnore`.

## Tests
- Add an Astro v7 + `@astrojs/cloudflare` adapter framework test (SSR + API env injection, log redaction, and verifying the wrangler `.env` message is replaced by the varlock notice).
- Force `astro dev` to run in the foreground in the test harness (`ASTRO_DEV_BACKGROUND`) — Astro v7 auto-backgrounds when it detects an agentic/CI environment, which detached output from the harness. This also fixed the pre-existing v7 node dev-server tests.
- `astro-v7.test.ts`: 33 passed, 3 skipped. Existing v5/v6 tests unaffected.

## Docs
- Add a "Deploying to Cloudflare Workers" section to the Astro integration docs, and list the Astro adapter in the Cloudflare integration's "Choosing an approach" table.

## Requirements for Astro + Cloudflare adapter
- Node ≥ 22
- `@astrojs/cloudflare` v14 requires Astro 7
- Don't add `varlockCloudflareVitePlugin()` alongside `@astrojs/cloudflare` — the adapter already includes `@cloudflare/vite-plugin`; a duplicate throws a conflict error


theoephraim:
does it make sense to move the cloudflare code into the vite integration?
I think you still want to tell users to use the cloudflare integration with the astro integration if they are using cloudflare?

theoephraim (approved):

PR #824, opened 2026-06-23
feat(varlock): honor _VARLOCK_* config vars set in .env/.env.local

theoephraim:
## What

Lets you set varlock's own config vars (`_VARLOCK_ENV_KEY`, `_VARLOCK_CACHE_KEY`, `_VARLOCK_REDACT_STDOUT`) as static literals in `.env` / `.env.local`, so they configure varlock during local development instead of being silently ignored. A real environment variable always takes precedence.

## Why

Setting `_VARLOCK_ENV_KEY` in a gitignored `.env.local` is a natural thing to want for local dev, but it previously had no effect (and the warning didn't explain why).

## Behavior

- **Honored only from `.env` / `.env.local`** — not the committed `.env.schema` or env-specific files. A `_VARLOCK_*` key elsewhere warns it has no effect.
- **Must be a plain literal** — these are read before the graph resolves, so a non-static value (`$ref`/`fn()`) is a hard error rather than a silent fallback.
- **Unrecognized `_VARLOCK_*` keys warn** (likely a typo).
- **They are varlock config, never app config** — excluded from the resolved env, `varlock load` output, the injected blob, generated types, and override provenance; not referenceable from other items; not forwarded to `varlock run` children (a real env var still is, via inheritance).
- `varlock cache` honors a file-set `_VARLOCK_CACHE_KEY` too.

## Notes

- Single read path for varlock's own config (`mergeVarlockConfigEnv` / `getVarlockConfigVar`), with `.env`-file values layered under the real environment.
- Reserved `_VARLOCK_*` keys are no longer registered as config items, which also fixes a prior inconsistency where they showed up in `load` output despite being excluded from the blob.

### DIFF
--- a/.bumpy/honor-varlock-config-vars-in-dotenv.md
+++ b/.bumpy/honor-varlock-config-vars-in-dotenv.md
@@ -1,0 +1,5 @@
+---
+varlock: minor
+---
+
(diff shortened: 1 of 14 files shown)


theoephraim:
failed experiment...

PR #841, opened 2026-06-26
fix(azure-key-vault): select cached CLI token by tenant

mskutle:
Hi friends!
I got an issue on my machine after I had switched between two Azure tenants. It appears like the plugin cached the access token I obtained from my first login (tenant A) when trying to access a keyvault in tenant B. 

getAzureCliToken() matched cached vault.azure.net tokens by scope and expiry only, never by tenant. With multiple `az login` accounts the MSAL cache can hold valid vault tokens for several tenants; the first match won regardless of which tenant issued it. Presenting a cross-tenant token to a Key Vault returns a 401 ("token expired or invalid").

Now, when a tenantId is configured:
- MSAL cache entries are filtered by `realm` (issuing tenant)
- legacy accessTokens.json entries are filtered via `_authority`
- the direct `az account get-access-token` fallback passes `--tenant`

All guards are no-ops when no tenantId is configured.

### DIFF
--- a/.bumpy/fix-azure-kv-cached-token-tenant.md
+++ b/.bumpy/fix-azure-kv-cached-token-tenant.md
@@ -1,0 +1,5 @@
+---
+'@varlock/azure-key-vault-plugin': patch
+---
+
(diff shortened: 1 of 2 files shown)


philmillman:
Thanks @mskutle we'll take a look today

theoephraim (approved):

PR #843, opened 2026-06-26
MVP array config item vals

philmillman:


packages/varlock-website/src/content/docs/reference/data-types.mdx
```diff
@@ -254,4 +254,34 @@ POLL_INTERVAL=15m
...
+
+**Element type** — first positional argument: a type name (`string`, `email`, …) or a nested call for types with positional args (`enum(dev, staging, prod)`).
+
+**Array options:**
+- `minLength` / `maxLength` — element count bounds
+- `unique` (boolean) — reject duplicate elements
+- `separator` (string) — split string input (e.g. `","` for comma-separated `.env` values)
+- `allowEmpty` (boolean) — allow `[]` (default: false)
+
+Named options not listed above are forwarded to the element type (e.g. `normalize=true` on `array(email, …)`).
```
theoephraim, line 271: this is weird

packages/varlock-website/src/content/docs/reference/data-types.mdx
```diff
@@ -254,4 +254,34 @@ POLL_INTERVAL=15m
...
+
+<div>
+### `array`
+
+Validates a list of values, coercing and checking each element with an inner type.
+
+**Element type** — first positional argument: a type name (`string`, `email`, …) or a nested call for types with positional args (`enum(dev, staging, prod)`).
+
+**Array options:**
+- `minLength` / `maxLength` — element count bounds
```
theoephraim, line 266: probable want an `isLength` too to match our string type

packages/varlock-website/src/content/docs/reference/data-types.mdx
```diff
@@ -254,4 +254,34 @@ POLL_INTERVAL=15m
...
+- `unique` (boolean) — reject duplicate elements
+- `separator` (string) — split string input (e.g. `","` for comma-separated `.env` values)
```
theoephraim, line 268: Both in terms of input and output, we may need some options around quotes and brackets.

For example, all these seem like reasonable inputs - `a,b,c`, `[a,b,c]`, `["a", "b", "c"]`
All of them could be parsed and the type-safe output will be fine, but the its not clear how we'd serialzie it back to a string for the basic env var. Also probably want to enable/disable basic splitting

Id assume default is that splitting by comma is fine, and no quotes needed, since if youre using an env var today to pass in a list, its probably just `a,b,c`

.bumpy/array-value-types.md
```diff
@@ -0,0 +1,7 @@
+---
+"@env-spec/parser": minor
+varlock: minor
+env-spec-language: minor
+---
+
+Add `@type=array(...)` for list-valued config items with per-element validation, array literal values, JSON and comma-separated input, and `T[]` type generation.
```
theoephraim, line 7: My debate internally had been whether to do wrap the type in `array()` or to split it into a new decorator --
`@type=string(startsWith=abc) @arrayType(minLength=2)`

Obviously wrapping is a bit more correct, but since our type stuff doesnt actually use normal function calls like the rest of the system I thought it might be cleaner. Also do we want to support nested arrays?

Im not set on it, but curious what you think.

Also - we may as well support object (record/dictionary) too with this work? I dont think arbitrary shaped objects like we had in dmno, but just setting the value type, and optionally the key type. 
`dictionary(uuid, keyType=string(isLength=2), minKeyCount=2)`

theoephraim:
Superseded by #917, which rebuilds this on current main (post-#892 emitter refactor) with object support, key validation, separator-symmetric serialization, and per-element redaction. Closing in favor of that PR.

PR #853, opened 2026-07-02
feat(cli): reject unknown/misspelled flags

theoephraim:
## What

gunshi silently ignores unrecognized options, so a typo like `varlock proxy rules --alow-reload` parsed to nothing and the command ran as if the flag were never passed, silently dropping the user's intent.

This adds a gunshi plugin that decorates every subcommand and rejects any option token that does not map to a declared arg, exiting non-zero with a "did you mean" suggestion:

```
Unknown flag: --alow-reload
Did you mean --allow-reload?
```

## How

- New `strict-flags-plugin.ts` uses `decorateCommand` to run a validation pass before each command.
- Only option tokens **before** the `--` terminator are checked, so child-process passthrough (`run`, `proxy run`) is untouched, e.g. `varlock run -- node --experimental-foo app.js` still works.
- Accepts declared `short` aliases, `--no-<name>` for negatable booleans, both `--flag=value` and `--flag value` forms, and always allows `--help`/`--version`.
- Dependency-free Levenshtein for the suggestion. Skips the internal `complete` command.

## Tests

Pure token-validation helpers are unit tested (unknown flags, short aliases, negatable, `--` passthrough, help/version, multi-unknown, suggestions), plus the `assertNoUnknownFlags` throwing path.

### DIFF
--- a/.bumpy/strict-cli-flags.md
+++ b/.bumpy/strict-cli-flags.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+Reject unknown or misspelled CLI flags with a did-you-mean suggestion instead of silently ignoring them
--- a/packages/varlock/src/cli/cli-executable.ts
+++ b/packages/varlock/src/cli/cli-executable.ts
@@ -2,6 +2,8 @@ import { cli, type Command } from 'gunshi';
 import completion from '@gunshi/plugin-completion';
 import { gracefulExit } from 'exit-hook';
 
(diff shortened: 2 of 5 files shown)


theoephraim:
Upstream feature request to move this into gunshi core (native `strict` / `onUnknownOption` + auto "did you mean"): kazupon/gunshi#611. If that lands, this plugin can be removed in favor of the built-in option.

PR #854, opened 2026-07-03
windows tpm support

philmillman:
## Summary

Adds hardware-backed at-rest key protection on Windows using NCrypt and the Platform Crypto Provider (TPM), while preserving Windows Hello and WSL behavior.

Previously, Windows local encryption was DPAPI plus an optional Hello presence gate — fingerprint auth did not mean the private key was TPM-bound. This change seals the P-256 scalar to the TPM (RSA-OAEP wrap via PCP), matching the Linux TPM2 pattern: TPM for at-rest storage, software ECIES for decrypt, Hello as a separate presence gate.

- New `ncrypt` protection tier — preferred when TPM/PCP is available; falls back to DPAPI, then plaintext
- `hardwareBacked: true` in `status` when NCrypt sealing works
- Auto-rewrap on decrypt — existing DPAPI keys upgrade to TPM sealing on the next successful decrypt (public key unchanged; no `.env` re-encryption)
- `rewrap-key` CLI — optional manual upgrade without decrypting
- Unattended note at `varlock encrypt` when hardware-backed but no presence gate (headless/CI)
- Hello + WSL unchanged — daemon, `UserConsentVerifier`, `--via-daemon`, PowerShell prestart, and `schtasks` fallback are untouched

## Design notes

- Hello is not replaced — still a pre-check in the daemon before unseal, not NGC/Passport KSP key-use integration (future work for Secure Enclave–like semantics)
- No downgrade — if a key is already `ncrypt` and TPM becomes unavailable, auto-rewrap will not rewrite it to DPAPI

## Test plan

- [x] Native Windows + TPM + Hello: `status` shows `hardwareBacked: true`; first decrypt prompts Hello; warm session ~5 min
- [x] Native Windows + TPM, no Hello: direct decrypt works unattended; `varlock encrypt` shows unattended note
- [x] Existing DPAPI key: decrypt once → key file becomes `protection: "ncrypt"`; encrypted values still decrypt
- [x] WSL2: decrypt via `--via-daemon` + fingerprint prompt unchanged
- [x] WSL2 cold start: daemon prestart / `schtasks` fallback still works
- [x] Machine without TPM: falls back to DPAPI (same as today)
- [x] `cargo test` in `packages/encryption-binary-rust`

### DIFF
--- a/.bumpy/windows-tpm-key-storage.md
+++ b/.bumpy/windows-tpm-key-storage.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+Windows local encryption now uses TPM-sealed keys via NCrypt when available; existing DPAPI keys auto-upgrade on the next decrypt.
--- a/packages/encryption-binary-rust/README.md
+++ b/packages/encryption-binary-rust/README.md
@@ -2,7 +2,7 @@
 
 Cross-platform local encryption binary for Varlock (Windows and Linux).
 
-Provides DPAPI key protection on Windows (with optional Windows Hello biometric), and TPM2 key protection on Linux, with a file-based plaintext fallback on both.
+Provides **NCrypt TPM-sealed** key protection on Windows when a TPM is available (with optional Windows Hello biometric), **DPAPI** as fallback, and TPM2 key protection on Linux, with a file-based plaintext fallback on both.
 
 ## Prerequisites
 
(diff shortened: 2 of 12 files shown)


philmillman:
ok I've tested this as extensively as I can, including disabling my TPM and disabling Hello and all combos thereof. I think we're good to go

theoephraim (approved):

PR #858, opened 2026-07-04
fix(vscode-plugin): don't highlight email-like text in trailing comments as decorators

theoephraim:
A trailing comment like `VAL=foo # contact email@asdf.com` highlighted `@asdf.com` with decorator annotation styling.

The `post-value-comment` TextMate rule included decorator styling unconditionally. It is now split into two variants: decorator styling only applies when the comment starts with `@`, mirroring the parser (which only treats comments starting with a decorator as decorator comments). Plain trailing comments — including email addresses and `@` mentions mid-text — render as regular comment text.

Adds a `.env.comments` snapshot fixture covering email-like text in standalone and trailing comments alongside real inline decorators.

### DIFF
--- a/.bumpy/fix-email-comment-decorator-highlight.md
+++ b/.bumpy/fix-email-comment-decorator-highlight.md
@@ -1,0 +1,5 @@
+---
+env-spec-language: patch
+---
+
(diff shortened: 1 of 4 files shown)


theoephraim:
will fix in the move to monogram instead

PR #870, opened 2026-07-08
feat(setValuesBulk): accept createMissing="sensitive" to create missing vars as sensitive

Jarva:
@setValuesBulk's createMissing option now accepts "sensitive" in addition to true/false. When set, keys created by the bulk source are marked sensitive regardless of @defaultSensitive, so their values are redacted in varlock load and other output instead of leaking as plain text.

### DIFF
--- a/.bumpy/set-values-bulk-create-missing-sensitive.md
+++ b/.bumpy/set-values-bulk-create-missing-sensitive.md
@@ -1,0 +1,5 @@
+---
+varlock: minor
+---
+
+@setValuesBulk createMissing option now accepts "sensitive" to create missing items as sensitive
--- a/packages/varlock-website/src/content/docs/reference/root-decorators.mdx
+++ b/packages/varlock-website/src/content/docs/reference/root-decorators.mdx
@@ -169,7 +169,7 @@ IMPORTED_ITEM=overridden-value
 <div>
 ### `@setValuesBulk()`
 **Arg types:** `[ data: string ]`
-**Named args:** `format?: "json" | "env"`, `createMissing?: boolean`, `enabled?: boolean`, `pick?: string[]`, `omit?: string[]`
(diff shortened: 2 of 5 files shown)


theoephraim:
Oh smart. I wonder if it should actually just be the default since usually you’ll be loading from vaults meant to hold secrets. 🤔

Jarva:
I considered that too but as a third-party contributor I decided it was easier to implement an option than change the default.

theoephraim:
@Jarva - just reviewing this and thinking through larger implications and design.
Question for you - did you consider just using `@defaultSensitive=true` in the file? Any new created items should respect that setting.

WalksWithASwagger (commented):
## Review notes (not approving yet — design still open with @theoephraim)

Implementation looks clean: validation, docs, bumpy minor, and a focused test for `createMissing="sensitive"` under `@defaultSensitive=false`.

On Theo’s question about `@defaultSensitive=true`: for the common “vaults hold secrets” case, that already covers newly created items without a new API. The unique value of `"sensitive"` is forcing redaction when the file’s default is *not* sensitive (exactly what the new test asserts).

Before merge, worth an explicit product call:
- keep the option (escape hatch when `@defaultSensitive=false`), or
- drop it and document “use `@defaultSensitive=true` with `createMissing=true`” instead.

Also note: `_forceSensitive` is set *after* `await newItem.process()`, so it relies on a later sensitivity pass / getter seeing the flag. Worth confirming that ordering is intentional and covered by the `expectSensitive` assertion path (not only `expectValues`).

I cannot merge this account (`WalksWithASwagger` has pull-only on the repo); leaving this for maintainers once the design call is settled.

theoephraim:
@Jarva - I think I will close this for now. Just relying on the file-level default should do the trick. We did have some discussion too around a [notion of "sections" or "blocks"](https://github.com/dmno-dev/varlock/discussions/149) which would let you set defaults and then have them apply just within one section. This will likley be a better path than adding individual flags to setValuesBulk

PR #872, opened 2026-07-08
feat: add sensitive/non-sensitive filter flag to load command

Jarva:
Adds the ability to filter `load` output by sensitive/non-sensitive variables. This allows easier automation with other tools like flyctl for Fly.io

`varlock load --format env --sensitive | fly secrets import`

### DIFF
--- a/.bumpy/load-sensitive-filter.md
+++ b/.bumpy/load-sensitive-filter.md
@@ -1,0 +1,5 @@
+---
+varlock: minor
+---
+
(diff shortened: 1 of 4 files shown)


theoephraim:
Thanks for this - have been wanting to add a new general `--filter` flag, so will roll selecting only sensitive stuff into that. Should have it out soon!

theoephraim:
closing in favor of #873

PR #873, opened 2026-07-08
Add --filter flag and @tag() decorator for selecting env vars

theoephraim:
## Summary
- Add `--filter` to `varlock load`/`run` for selecting/omitting env vars by key name, glob (`STRIPE_*`), negation (`!KEY`), decorator (`@sensitive`/`@required`), or tag (`#tagname`)
- Add a new `@tag()` item decorator to attach tags to items, used by `#tagname` selectors. Tag names are validated (start with a letter/number, then letters/numbers/`_`/`-`) so every tag stays selectable via `#tagname`; invalid names and empty `@tag()` calls are schema errors, and duplicates collapse
- Add a matching `filter=` arg to `@generate*` code-generation decorators (`@generateTsTypes`, `@generatePythonEnv`, etc.), so one schema can emit multiple generated files scoped to different subsets (e.g. a billing-only `env.d.ts`)
- On `run`, a filtered-out var can't reach the child process at all: excluded items are stripped from the `__VARLOCK_ENV` blob (including its override-provenance metadata) and removed from the child env even when set ambiently (same treatment as `@internal`). Without the ambient stripping, `SECRET=x varlock run --filter='!@sensitive' -- ...` passed the ambient value through `process.env` and, since excluded items are left out of the redaction map, printed it unredacted
- Add a `_VARLOCK_FILTER` env var fallback for `--filter`, for wrapper scripts, CI config, or build-time integrations that don't go through the CLI (an explicit `--filter` flag still takes precedence)
- **A key/glob/tag-only `--filter` now scopes resolution and validation too**, not just output: `load`/`run` resolve only the selected items plus their transitive dependencies, so an unrelated broken item outside the filter no longer blocks the load - e.g. a build step scoped to `--filter="#frontend"` doesn't care if a backend-only var is misconfigured. `@sensitive`/`@required` selectors can't be scoped this way (matches aren't knowable until the graph is resolved), so those still resolve and validate everything, same as before.
- Docs spell out how selectors of different kinds combine: non-negated selectors OR together regardless of kind, then negated selectors subtract from that whole set (no intersection is expressible)
- A `--filter` that matches no items (e.g. a typo'd tag) warns on stderr; the command still succeeds and stdout stays clean for piping
- VSCode extension: autocomplete/hover entries for `@tag()`, the `filter=` arg on `@generate*` decorators, and `@internal` (pre-existing gap)

Credit to @Jarva for #872, which proposed `--sensitive`/`--non-sensitive` flags on `load` for piping secrets to tools like `fly secrets import`. This PR generalizes that idea into a single `--filter` flag that also covers `@required`, tags, and key/glob selection.

Fixed along the way: the codegen `filter=` arg initially evaluated `@sensitive`/`@required` against raw `ConfigItem` getters, which aren't populated until `resolveEnvValues()` runs - but code generation runs before that, so those selectors silently matched everything. Filtering now goes through the same pre-resolution computation (`getTypeGenInfo()`) codegen already uses for `isSensitive`/`isRequired`.

Rebased on top of #874 (merged), which fixes `load --format json-full` to exclude `@internal` items by default - discovered while documenting `--filter`'s interaction with `@internal`. `load.command.ts`'s `json-full` branch now combines both fixes (`includeInternal` + `filterKeys`); verified end-to-end that `--include-internal` and `--filter` compose correctly together.

## Test plan
- [x] `bun run --filter varlock typecheck`
- [x] `bunx vitest run` in `packages/varlock` (1000 tests passing, including coverage for the filter parser, `@tag()` decorator, codegen `filter=` wiring, mixed-kind selector interaction, `@internal` exclusion holding regardless of filter, the `_VARLOCK_FILTER` env var/precedence, scoped resolution/validation for key/glob/tag filters, and `getSerializedGraph` filterKeys/override-provenance exclusion), plus smoke tests covering `run --filter` ambient-env stripping
- [x] `bun run lint:fix`
- [x] `bun run --filter varlock-website astro build`
- [x] Manually verified `varlock load --filter=...`, `varlock run --filter=...`, `varlock codegen` with mixed `filter="#tag,@sensitive"`, `_VARLOCK_FILTER` (plus flag precedence), scoped resolution (an unrelated invalid item outside a key/tag filter no longer fails the load, while `--filter=@sensitive` still catches it), and `--include-internal` + `--filter` composed together end-to-end against real `.env.schema` files


philmillman (approved):
lgtm, let's consider an in-depth tags and filters guide for the future

PR #877, opened 2026-07-11
Add @excludeSensitiveFromInjectedEnv root setting

### LINKED ISSUE #876: [Feature Request] Exclude @sensitive values from the resolved-env blob when the runtime provides them
### Description

When using `@varlock/vite-integration` with `ssrInjectMode: 'resolved-env'`, the injected `globalThis.__varlockLoadedEnv` blob contains the plaintext value of every `@sensitive` item. I'd like a root setting (`@excludeSensitiveFromInjectedEnv`, symmetric to `@encryptInjectedEnv`) that keeps those values out of the bundle-embedded blob, with `ENV` falling back to `process.env` at runtime for the excluded items.

### Motivation

The vite docs already note the blob is plaintext and that anyone with access to the build artifact can read the secrets. On some platforms the artifact is more exposed than runtime env: Cloudflare worker script contents are readable through the API by anyone with account read access, and archived per script version, while runtime secret bindings are write-only. `varlock-wrangler` itself uploads `__VARLOCK_ENV` as write-only secrets (chunked, streamed via FIFO so values never touch disk), and that runtime binding overrides the embedded blob anyway. So on that path, and for any app that reads secrets from platform runtime env (e.g. SvelteKit's `$env/dynamic/private`), the embedded plaintext values are never read and are pure at-rest exposure.

I hit this in production: build-time-only deploy credentials were readable in a deployed worker script even though nothing at runtime read them from the blob.

The framework tests assert "sensitive value is absent from all output" for the default inject mode, sourcemaps, HTML replacement and log redaction. Unencrypted `resolved-env` is the one configuration where that invariant does not hold today.

### Proposed Solution

A root setting `@excludeSensitiveFromInjectedEnv`:

- the vite integration (build only, dev unaffected) serializes sensitive items into the blob without `value`, flagged `valueExcluded: true`
- `initVarlockEnv()` reads excluded items from `process.env` and leaves `process.env` untouched for them (today an undefined value is re-injected as `''`, which would clobber a platform-provided runtime value)
- composes with `@encryptInjectedEnv` (strip first, then encrypt)

PR incoming.

### Alternatives

- `@encryptInjectedEnv`: protects the blob, but needs key management, and the values are still inside the artifact for anyone holding the key
- `@internal` (#817): right for vars the app never needs, but it also strips them from the `varlock run` child env, so it does not fit vars consumed by build scripts or supplied by the platform at runtime
- what I run today: a custom vite plugin that post-processes the manifest before it is serialized, which means reimplementing integration internals downstream

### Any additional comments?

Found while auditing what a deployed SvelteKit + Cloudflare worker bundle actually contains.

stickerdaniel:
## What

New root setting `@excludeSensitiveFromInjectedEnv`. When enabled, the vite integration's `resolved-env` inject mode writes `@sensitive` items into the bundle-embedded blob without their values, flagged `valueExcluded: true`, and `initVarlockEnv()` falls back to `process.env` for those keys. The fallback also leaves `process.env` untouched for excluded items, so a platform-provided runtime value is never clobbered with `''`. Dev mode keeps values embedded (the dev bundle is not a deploy artifact, mirroring how the encryption path special-cases dev), and the setting composes with `@encryptInjectedEnv` (strip first, then encrypt).

The strip lives in the integration rather than `getSerializedGraph()` because other serialization channels legitimately need the values (`varlock run` child env, the varlock-wrangler secret binding, `load --format json-full` inspection). Only the bundle-embedded copy must not carry them.

A separate commit adds a one-time notice when a production `resolved-env` build embeds unencrypted sensitive values. Easy to drop if unwanted.

## Why

The resolved-env blob is plaintext in the deploy artifact (the vite docs call this out). When the runtime already supplies sensitive values (platform env, Cloudflare bindings, or the varlock-wrangler `__VARLOCK_ENV` secret binding that overrides the embedded blob at runtime anyway), the embedded plaintext values have no consumer and are pure at-rest exposure. On Cloudflare, worker script contents are API-readable and archived per version, unlike the write-only secret bindings varlock-wrangler uses.

Resolves #876

New tests: runtime env-state tests for the `process.env` fallback and no-clobber behavior, and a vite framework-test scenario asserting the sensitive value is absent from SSR output while public values and the blob remain. Docs updated (vite integration, root decorators reference, encrypted deployments guide).

### DIFF
--- a/.bumpy/exclude-sensitive-from-injected-env.md
+++ b/.bumpy/exclude-sensitive-from-injected-env.md
@@ -1,0 +1,6 @@
+---
+"varlock": minor
+"@varlock/vite-integration": minor
+---
+
+new @excludeSensitiveFromInjectedEnv root setting keeps @sensitive values out of the SSR bundle injected by resolved-env mode; ENV falls back to runtime process.env for excluded items
--- a/framework-tests/frameworks/vite/files/schemas/.env.schema.exclude-sensitive
+++ b/framework-tests/frameworks/vite/files/schemas/.env.schema.exclude-sensitive
@@ -1,0 +1,14 @@
+# @defaultSensitive=false @defaultRequired=infer
+# @currentEnv=$APP_ENV
+# @excludeSensitiveFromInjectedEnv=true
+# ---
+
+# @type=enum(dev, prod)
+APP_ENV=dev
+
+PUBLIC_VAR=public-test-value
+API_URL=https://api.example.com
+ENV_SPECIFIC_VAR=env-specific-default
+
+# @sensitive
+SECRET_KEY=super-secret-value
--- a/framework-tests/frameworks/vite/vite-shared.ts
+++ b/framework-tests/frameworks/vite/vite-shared.ts
@@ -289,6 +289,33 @@ export function defineViteTests(
       });
     });
 
+    // ---- Exclude sensitive from injected env ----
+
+    describe('exclude sensitive from injected env', () => {
+      viteEnv.describeScenario('SSR build strips @sensitive values from the resolved-env blob', {
+        command: 'vite build --ssr src/ssr-entry.ts',
+        templateFiles: {
+          'vite.config.ts': 'vite-configs/vite.config.resolved-env.ts',
+          '.env.schema': 'schemas/.env.schema.exclude-sensitive',
+          'index.html': 'html/basic.html',
+          'src/ssr-entry.ts': 'pages/ssr-entry.ts',
+        },
+        expectSuccess: true,
+        fileAssertions: [
+          {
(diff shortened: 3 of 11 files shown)


philmillman:
Thanks for this @stickerdaniel. We recently added generic filter functionality which might cover this but I'll let @theoephraim comment

stickerdaniel:
The key filter removes the item from the blob entirely -> ENV.KEY throws at runtime

This is a solution that keeps the key and drops only its value so the runtime falls back to process.env for a value provided by the platform

theoephraim:
hop in our discord (if you're not there already) and let's talk through this. I think your issues can be solved without introducing new behaviour.

PR #878, opened 2026-07-11
allowMissing for Infisical

### LINKED ISSUE #875: [Feature Request] `allowMissing` for Infisical plugin (`infisical()` / `@initInfisical`)
### Description
Please add an `allowMissing` flag to `@varlock/infisical-plugin`, mirroring the behavior already shipped for 1Password in [#586](https://github.com/dmno-dev/varlock/pull/586).

Today, `infisical()` always throws a `ResolutionError` when Infisical returns 404 / “not found”. That makes optional Infisical-backed secrets impossible to express cleanly with Varlock’s normal composition tools (`fallback()`, `@required=false`, `forEnv()`, etc.).

**Environment**

- `@varlock/infisical-plugin@2.0.1`
- `varlock@^1.10.0`
- Infisical Cloud (Universal Auth machine identity)

**Repro shape**

```env-spec
# @currentEnv=$APP_ENV
# @defaultRequired=false @defaultSensitive=false
# @plugin(@varlock/infisical-plugin)
# @initInfisical(
#   projectId=$INFISICAL_PROJECT_ID,
#   environment=remap($APP_ENV, "development", dev, "staging", staging, "production", prod, "test", dev),
#   clientId=$INFISICAL_CLIENT_ID,
#   clientSecret=$INFISICAL_CLIENT_SECRET
# )
# ---

# @sensitive
RESEND_API_KEY=infisical()

# @sensitive
OPENAI_API_KEY=infisical()
```

**Observed behavior**

1. If those keys are **absent** from the Infisical environment, `varlock load --agent` fails with `ResolutionError: Secret "…" not found`.
2. Marking the item optional (`@defaultRequired=false`, no `@required`) does **not** help — the failure happens during resolver execution, before optional-value validation can accept `undefined`.
3. `fallback(infisical(), undefined)` / `fallback(infisical(), "")` also does **not** help, for the same reason as pre-`allowMissing` 1Password: the inner resolver throws instead of returning empty/undefined.

In `@varlock/infisical-plugin@2.0.1`, missing secrets are explicitly converted into resolution failures (404 / NotFound → `ResolutionError`). There is currently no `allowMissing` argument on `infisical()` or `@initInfisical`.

### Motivation
We have a large schema where many secrets are **declared** but only **required when a feature is enabled** (email provider, AI provider, Stripe live mode, calendar, Alchemy, etc.).

Without `allowMissing`, teams are pushed into non-idiomatic workarounds:

1. **Seed sentinel values** into Infisical for every optional key (`UNSET`, etc.) so `infisical()` always succeeds — pollutes the secret store and forces app/runtime code to treat sentinels as empty.
2. **Omit `infisical()`** for optional keys and keep dual paths (schema empty + sometimes Infisical) — fights an “Infisical owns secrets” model.
3. **Over-require secrets** in every environment — hurts local/dev/test ergonomics.

This matches the exact motivation in [#586](https://github.com/dmno-dev/varlock/pull/586):

> `op()` had no way to signal that a missing item was acceptable — it always threw, making `fallback(op(...), "default")` unusable when the 1Password item doesn't exist.

`allowMissing` would restore the Varlock-native composition model already documented for 1Password.

### Proposed Solution
Parity with 1Password’s `allowMissing` for `op()` / `@initOp()`:

**Per-call**

```env-spec
# If present in Infisical, use it; if missing, resolve to undefined and fall through
RESEND_API_KEY=fallback(infisical(allowMissing=true), undefined)

# Or with an explicit default
SOME_OPTIONAL_KEY=fallback(infisical(allowMissing=true), "")
```

**Instance-level default**

```env-spec
# @initInfisical(
#   projectId=$INFISICAL_PROJECT_ID,
#   environment=dev,
#   clientId=$INFISICAL_CLIENT_ID,
#   clientSecret=$INFISICAL_CLIENT_SECRET,
#   allowMissing=forEnv(development, test)
# )
```

**Semantics (matching #586)**

When `allowMissing` is truthy:

- Infisical **not found** (404 / NotFound) → return `undefined` (do not throw)
- Auth failures, permission errors, network/service errors → **still throw**
- Invalid config / bad instance id → **still throw**

Support dynamic values (`allowMissing=forEnv(development)`, `allowMissing=true`, etc.), not only static literals. Per-call `allowMissing` should override the instance default.

Docs: keep default fail-closed “Secret not found” troubleshooting, and additionally document the optional-secret + `fallback()` pattern (same as the 1Password plugin README).

### Alternatives
1. **Keep seeding Infisical with sentinel values** (`UNSET`) for every optional schema key — works, but pollutes the secret inventory and leaks a workaround into application empty-checks.
2. **Only call `infisical()` when a feature flag/env requires it** via `if()` / `forEnv()` wrappers — possible for some keys, but awkward for many optional provider secrets and still fails if the feature is on and the secret is absent.
3. **Leave optional secrets as empty schema values** (no Infisical resolver) until needed — reintroduces dual-path secret ownership.
4. **Wait / contribute upstream** — preferred; this request is specifically for Infisical parity with the existing 1Password solution rather than a new abstraction.

### Additional Information
- Precedent: [#586](https://github.com/dmno-dev/varlock/pull/586) (`allowMissing` for `op()` / `@initOp()`)
- Plugin docs: https://varlock.dev/plugins/infisical/
- Package: `@varlock/infisical-plugin`
- Happy to validate a prerelease build against a real Infisical Cloud project (Universal Auth + multi-env remap: `development/test→dev`, `staging→staging`, `production→prod`).

philmillman:
## Summary

Adds `allowMissing` to `@varlock/infisical-plugin`, matching the behavior already shipped for 1Password in #586.

Closes #875.

When `allowMissing` is truthy, Infisical 404 / not-found errors resolve to `undefined` instead of throwing. That makes optional Infisical-backed secrets work with Varlock's normal composition tools (`fallback()`, `@required=false`, `forEnv()`, etc.). Auth failures, permission errors, and other non-not-found errors still throw.

- **`infisical(allowMissing=true)`** — per-call override
- **`@initInfisical(allowMissing=...)`** — instance-level default (supports dynamic values like `forEnv(development)`)
- Per-call `allowMissing` overrides the instance default

## Example

```env-spec
# @initInfisical(
#   projectId=$INFISICAL_PROJECT_ID,
#   environment=dev,
#   clientId=$INFISICAL_CLIENT_ID,
#   clientSecret=$INFISICAL_CLIENT_SECRET
# )
# ---

# Optional provider secret: use Infisical when present, otherwise undefined
RESEND_API_KEY=fallback(infisical(allowMissing=true), undefined)

### DIFF
--- a/.bumpy/infisical-allow-missing.md
+++ b/.bumpy/infisical-allow-missing.md
@@ -1,0 +1,5 @@
+---
+"@varlock/infisical-plugin": minor
+---
+
+Add allowMissing flag to infisical() and @initInfisical() for optional secrets
--- a/bun.lock
+++ b/bun.lock
@@ -314,6 +314,7 @@
         "@env-spec/utils": "workspace:^",
         "@infisical/sdk": "^5.0.2",
         "@types/node": "catalog:",
+        "outdent": "catalog:",
         "tsup": "catalog:",
         "varlock": "workspace:^",
         "vitest": "catalog:",
--- a/packages/plugins/infisical/README.md
+++ b/packages/plugins/infisical/README.md
@@ -14,6 +14,7 @@ Load secrets from [Infisical](https://infisical.com/) into your Varlock configur
 - ✅ Secret paths and hierarchical organization
 - ✅ Filter secrets by tag
 - ✅ Multiple plugin instances for different projects/environments
+- ✅ **`allowMissing`** option for optional secrets
 - ✅ Helpful error messages with resolution tips
 
 ## Installation
@@ -93,6 +94,7 @@ See the [OIDC Workload Identity guide](https://varlock.dev/guides/oidc/) for ful
 - **`siteUrl`** (optional): Custom Infisical instance URL (defaults to `https://app.infisical.com`)
 - **`secretPath`** (optional): Default secret path for all secrets (defaults to `/`)
 - **`cacheTtl`** (optional): Cache resolved values for the specified duration (`"5m"`, `"1h"`, `"1d"`, or `"forever"` to cache until manually cleared); set to `false` (or an empty string) to disable caching
+- **`allowMissing`** (optional): When `true`, missing secrets return `undefined` instead of throwing. Can be a dynamic value (e.g., `allowMissing=forEnv(development)`). Other errors (auth failures, permission errors, etc.) still throw.
 - **`id`** (optional): Instance identifier for using multiple instances
 
 ## Usage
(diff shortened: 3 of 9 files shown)


theoephraim:
lgtm - we may want to think about solving this in a more generalized way rather than per plugin, but it might be hard to do that... So likely just doing it per plugin for now

theoephraim (approved):

PR #882, opened 2026-07-13
Docs/audit round2

philmillman:
more feedback from https://docs.review/reports/01KXDZ1W269ENMZFN8FYAHH83Y

I left out the plugin initialization on the homepage since it made it a bit too long


philmillman:
@theoephraim main change here is dropping `encrypted()`, just let me know that this makes sense

PR #883, opened 2026-07-13
feat(nextjs): add disableWatch option to skip extra env-file watching

mhornbacher:
## Summary
- Adds `disableWatch` to `varlockNextConfigPlugin` (sets `process.env.__VARLOCK_NEXT_DISABLE_WATCH`)
- Next forwards that delta into `@next/env` `updateInitialEnv`, which tears down extra-file watchers
- Watched paths are shared across duplicate CJS copies of `next-env-compat` via `globalThis` + `process.env.__VARLOCK_NEXT_WATCHED_EXTRA_FILES` (loadEnvConfig and updateInitialEnv can hit different copies)
- Unit tests cover same-instance and cross-instance teardown

## Test plan
- [ ] Rebuild/relink `@varlock/nextjs-integration`
- [ ] `DEBUG_VARLOCK_NEXT_INTEGRATION=1` with `disableWatch: true`
- [ ] Expect `next-env-compat module init <id>` (possibly two ids), then after plugin: `updateInitialEnv` with `{ instance, newEnv }` and `disableWatch: unwatchAllExtraFiles { instance, count, paths }`
- [ ] FIFO `.env` watch loop stops
- [ ] `bun run --filter @varlock/nextjs-integration test:ci`

### DIFF
--- a/.bumpy/nextjs-disable-watch.md
+++ b/.bumpy/nextjs-disable-watch.md
@@ -1,0 +1,5 @@
+---
+"@varlock/nextjs-integration": minor
+---
+
+Add disableWatch option to varlockNextConfigPlugin to skip extra env-file watching (useful when sources are FIFOs).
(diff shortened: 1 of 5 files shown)


theoephraim:
I think I solved this with a different tactic - we can just detect the FIFO and not watch it at all.
This still leaves watching/reloads intact when you edit other files.

PR #889, opened 2026-07-17
fix env spec expansion

philmillman:
Root cause: Post-parse expansion in @env-spec/parser, not the grammar or exec runtime.

Changes:

- Nested parens — Replaced /\$\(([^)]+)\)/g with a quote-aware balanced-paren scanner in [expand.ts](vscode-file://vscode-app/Applications/Cursor.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/packages/env-spec-parser/src/expand.ts)
- Space stripping — Added skipTrim on ParsedEnvSpecStaticValue and staticFromExpansionSlice() so ref splits keep adjacent whitespace
- Tests — Regression coverage for nested ), quoted ), and spaces before $HOST / ${HOST} (248 passing)
- Test harness — simpleResolver now resolves nested concat/ref inside exec args (matching real ExecResolver)

### DIFF
--- a/.bumpy/fix-exec-expansion-nested-parens.md
+++ b/.bumpy/fix-exec-expansion-nested-parens.md
@@ -1,0 +1,5 @@
+---
+"@env-spec/parser": patch
+---
+
+Fix $(...) truncation on nested parentheses and dropped spaces before $VAR/${VAR} inside unquoted exec expansion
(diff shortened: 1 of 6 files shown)


philmillman:
@theoephraim there might be a simpler way than `findExecExpansions`

theoephraim (approved):
🚀

PR #895, opened 2026-07-18
Add WSL detection to install script

cturner8:
When running the install script on WSL, the `varlock-local-encrypt.exe` binary  was ignored meaning anything checking `isWSL` won't find the expected exe

The linux block to install the local encrypt binary now also copies the `exe`, unless overriden with new `--skip-win-exe ` flag

Installation folder now contains the linux native binary and windows exe:

image

`varlock encrypt` now correctly resolves the Windows TPM backend:

image

### DIFF
--- a/.bumpy/fix-wsl-install-encrypt-exe.md
+++ b/.bumpy/fix-wsl-install-encrypt-exe.md
@@ -1,0 +1,5 @@
+---
+"varlock": patch
+---
+
+install.sh now installs varlock-local-encrypt.exe on WSL so local encryption can use the Windows TPM/Hello backend (pass --skip-win-exe to opt out)
--- a/packages/varlock-website/src/content/docs/getting-started/installation.mdx
+++ b/packages/varlock-website/src/content/docs/getting-started/installation.mdx
@@ -69,11 +69,11 @@ Usage: install.sh [options]
 install varlock binary
 
 Options:
-  --dir             directory to install varlock to (default: "$HOME/.config/varlock/bin")
+  --dir             directory to install varlock to (defaults to $XDG_CONFIG_HOME/varlock/bin, else ~/.varlock/bin if it exists, else ~/.config/varlock/bin)
   --reinstall       reinstall even if already installed (default: false)
   --version         version of varlock to install (defaults to latest)
   --force-no-brew   force install without homebrew even when detected (default: false)
-  --skip-win-exe   skip installing the Windows native encryption binary, only applicable when running in WSL (default: false)
+  --skip-win-exe    on WSL, skip installing the Windows encryption helper (varlock-local-encrypt.exe) (default: false)
 ```
 
 ## Varlock skill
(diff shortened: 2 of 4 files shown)


philmillman:
Thank you @cturner8, will take a look soon!

WalksWithASwagger (dismissed):
## Verdict: needs changes before merge

`is_wsl` itself looks fine and matches `wsl-detect.ts` (`WSL_DISTRO_NAME` → `/proc/version` → `uname -r`).

The linux install branch is the problem. Current head:

```sh
if is_wsl && [ "$FORCE_LINUX_ENC_BIN" = "false" ] && [ -f "${_temp_dir}/varlock-local-encrypt.exe" ]; then
  # install .exe AND linux binary
fi
```

That regresses non-WSL Linux: the native encrypt binary is never installed. It also makes `--force-linux-enc-bin` a no-op / disable (condition fails, nothing installs).

Suggested shape:

1. Always install `varlock-local-encrypt` when present (non-WSL Linux + WSL fallback).
2. On WSL, also install `varlock-local-encrypt.exe` when present (unless `--force-linux-enc-bin`).
3. Keep `--force-linux-enc-bin` as “skip the `.exe` path on WSL,” not “skip all encrypt binaries.”

Also guard each `install` with its own `-f` check so a missing companion binary does not fail the whole step.

Happy to re-review after that split.

WalksWithASwagger:
Opened an alternate PR with the same goal but without regressing non-WSL Linux installs: the Linux helper always installs when present; WSL additionally gets `varlock-local-encrypt.exe` unless `--skip-win-exe`.

See: https://github.com/dmno-dev/varlock/pulls?q=is%3Apr+head%3Afix-install-wsl-encrypt-exe

WalksWithASwagger:
Update: the corrected PR is https://github.com/dmno-dev/varlock/pull/896 (same goal; Linux helper always installs, WSL additionally gets the Windows `.exe` unless `--skip-win-exe`).

philmillman:
there's obviously some duplication between this and #896, it would be good to consolidate. Let us know if either of you feel strongly either way.

cturner8:
> there's obviously some duplication between this and #896, it would be good to consolidate. Let us know if either of you feel strongly either way. 

#896 looks to fully replicate the updated of install.sh from this branch, mainly looks to add docs updates which I'd be happy to update here

philmillman:
> > there's obviously some duplication between this and #896, it would be good to consolidate. Let us know if either of you feel strongly either way.
> 
> #896 looks to fully replicate the updated of install.sh from this branch, mainly looks to add docs updates which I'd be happy to update here

works for me!

cturner8:
> > > there's obviously some duplication between this and #896, it would be good to consolidate. Let us know if either of you feel strongly either way.
> > 
> > 
> > #896 looks to fully replicate the updated of install.sh from this branch, mainly looks to add docs updates which I'd be happy to update here
> 
> works for me!

updated

PR #896, opened 2026-07-19
fix(varlock): install Windows encrypt helper on WSL via install.sh

WalksWithASwagger:
## Summary
- Linux standalone archives already bundle `varlock-local-encrypt.exe` for WSL2; `install.sh` now installs it when WSL is detected.
- Non-WSL Linux still always gets `varlock-local-encrypt` (fixes the regression in #895's current approach).
- Adds `--skip-win-exe` to opt out of the Windows helper on WSL.
- Documents the behavior in the local encryption guide.

Related: #895 (same intent; this avoids gating the Linux helper on WSL/`.exe` presence)

## Test plan
- [x] `sh -n packages/varlock/install.sh`
- [x] `is_wsl` false on macOS; true when `WSL_DISTRO_NAME` is set
- [ ] On WSL: `curl -sSfL … | sh -s -- --force-no-brew` installs both helpers; `varlock encrypt` resolves Windows TPM backend
- [ ] On non-WSL Linux: Linux helper still installed; no `.exe` required
- [ ] On WSL with `--skip-win-exe`: Linux helper only

### DIFF
--- a/.bumpy/fix-install-wsl-encrypt-exe.md
+++ b/.bumpy/fix-install-wsl-encrypt-exe.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+Install `varlock-local-encrypt.exe` from the Linux standalone archive when running on WSL, so local encryption can use the Windows TPM/Hello backend.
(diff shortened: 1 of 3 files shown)


theoephraim:
@WalksWithASwagger - is this something you encountered, tried out the other PR, and then came to this? Or is it just your clanker looking for stuff to contribute?

theoephraim:
Combined the changes into #895 - should be rolled out soon.
In future, feel free to leave specific comments on open PRs (not reviews) - but try to keep it targeted/minimal to specific things you've encountered. Just harder to sort through large volume and lots of text.

Thanks!

PR #912, opened 2026-07-20
fix: JSON-encode object values for process.env injection

### LINKED ISSUE #900: [BUG]: simple-object values become "[object Object]" in process.env / varlock run
## Problem

Coerced non-string values (notably `@type=simple-object`) are destroyed when injected into `process.env` / child process env via `String(itemValue)`, which produces `"[object Object]"`.

`varlock load --format env` already JSON-encodes objects correctly, so the intended encoding exists on one path but not the common injection paths.

## Evidence

`packages/varlock/src/runtime/env.ts` (`initVarlockEnv`):

```ts
process.env[itemKey] = itemValue === undefined ? '' : String(itemValue);
```

`packages/varlock/src/cli/commands/run.command.ts` spreads resolved values into spawn `env` similarly.

Contrast: `packages/varlock/src/cli/commands/load.command.ts` uses `JSON.stringify` for object values in `--format env`.

## Minimal repro

```env-spec
# ---
OBJ={"a":1}  # @type=simple-object
```

Then run `varlock run -- node -e 'console.log(process.env.OBJ)'` (or any path that calls `initVarlockEnv` with `setProcessEnv`).

Expected: `{"a":1}` (JSON string), matching `load --format env`.
Actual: `[object Object]`.

## Notes

- No injection round-trip test for `simple-object` today.
- Same footgun likely affects other non-string coerced types when they land in `process.env`.

philmillman:
## Summary
- Fixes #900. Objects became "[object Object]" via String(). Shared helper JSON-encodes for process.env and varlock run, matching load --format env.

## Test plan
- [ ] Confirm object env values round-trip as JSON
- [ ] Run related unit/smoke tests

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.bumpy/fix-simple-object-process-env.md
+++ b/.bumpy/fix-simple-object-process-env.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+JSON-encode object env values for process.env and varlock run
--- a/packages/varlock/src/cli/commands/load.command.ts
+++ b/packages/varlock/src/cli/commands/load.command.ts
@@ -4,6 +4,7 @@ import { gracefulExit } from 'exit-hook';
 
 import { loadVarlockEnvGraph } from '../../lib/load-graph';
 import { getItemSummary } from '../../lib/formatting';
+import { envValueToProcessEnvString } from '../../lib/env-value-to-string';
 import { redactString } from '../../runtime/lib/redaction';
 import {
   checkForConfigErrors, checkForNoEnvFiles, checkForSchemaErrors, showPluginWarnings,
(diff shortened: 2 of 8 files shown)


theoephraim:
Heads up: #917 now covers this fix as part of the array/record value-types work. All the injection sites patched here (run, proxy, load --format env, runtime re-injection) flow through type-owned serialization there: each data type controls its process.env string form, so simple-object serializes as JSON (and typed arrays get separator-joined forms). Tests covering the #900 scenario are included. Once #917 lands this can probably be closed.

PR #927, opened 2026-07-22
mvp benchmarking - WIP

philmillman:
Adds a release benchmarking suite under `benchmarks/` measuring memory footprint, execution time, and the latency added by redaction and leak prevention. It runs against published npm packages (plus the linux SEA binary when available), triggered after a varlock publish or manually.

## What it covers

CLI (`load` cold/warm, `run` wrap overhead, stdout redaction, `scan`/`audit`), the Next and Vite integrations (build cost against a no-varlock baseline, plus request latency for `preventLeaks` and `redactLogs`), and Python/Go codegen. CLI scenarios are measured across three runtimes: npm installed and run under node, bun installed and run under bun, and the compiled standalone binary.

Results are written to `results/` as JSON and committed by CI so trends show up in git history. Each run prints a deltas table (varlock vs baseline, telemetry on vs off, redaction on vs off), and any delta smaller than the spread of its inputs is tagged as noise rather than reported as a change. Absolute wall times on a shared runner are not comparable between runs; the within-run deltas are.

The integration benches reuse the Next/Vite templates from `framework-tests` via a new `usePublished` mode on `FrameworkTestEnv`, which installs from npm instead of packing workspace tarballs. The imperative parts of that harness moved to `harness/fixture-env.ts` so it can be used without loading Vitest.

## Telemetry

Telemetry-on scenarios measure what the telemetry code path costs, but never send real telemetry. `CONFIG.POSTHOG_HOST` is now overridable through `VARLOCK_POSTHOG_HOST`, and the suite points it at a local mock collector. Before running those scenarios it probes whether the version under test honours the override, and skips them with a recorded note if not, so synthetic events cannot reach product analytics.

This is not an opt-out knob: `VARLOCK_TELEMETRY_DISABLED` still disables telemetry entirely.

Those scenarios stay skipped until the first release that includes the config change.

## Known gaps

- Linux/x64 only, so the macOS and Windows binaries are not measured
- The `cli-load` fixture is all static literals, so cold vs warm has little to distinguish. Exercising the cache properly needs a fixture with expensive resolvers
- Informational only: no regression gate, and nothing reads the accumulated `results/` history yet

`integration-next` has not been run end to end yet. The other groups have.

### DIFF
--- a/.bumpy/bench-telemetry-host-override.md
+++ b/.bumpy/bench-telemetry-host-override.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+Allow overriding the telemetry endpoint with VARLOCK_POSTHOG_HOST, so tooling can point it at a local mock
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,0 +1,152 @@
+name: Benchmarks
+
+# Runs against published npm packages (+ optional SEA binary). Triggered after a
+# varlock publish (via release.yaml) or manually for iteration.
+on:
+  workflow_dispatch:
+    inputs:
+      varlock_version:
+        description: 'Published varlock version (empty = latest on npm)'
+        required: false
+        type: string
+        default: ''
+      only:
+        description: 'Optional comma-separated scenario groups (empty = all)'
+        required: false
+        type: string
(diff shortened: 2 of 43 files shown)


philmillman:
@pullfrog review this

PR #936, opened 2026-07-24
Fix proxy certs for strict TLS verifiers + 1password windows support

theoephraim:
Two fixes from user feedback on the proxy and 1password plugin.

## Proxy: minted MITM certs rejected by strict TLS verifiers

Python 3.13+ enables `VERIFY_X509_STRICT` by default, and our minted CA/leaf certs had no subject/authority key identifier extensions, so every modern python client (urllib and httpx alike) failed the TLS handshake with `Missing Authority Key Identifier`. The CA cert now carries an SKI and minted leaves carry SKI + AKI (derived from the CA public key so it byte-matches the CA's SKI).

Also in this area:
- Client-sent `Proxy-Authorization` is now stripped (hop-by-hop, addressed to the proxy) instead of forwarded to the destination host.
- New smoke test drives real python urllib (with strict verification forced) through `proxy run` via both the installed CLI and the compiled binary. Verified it catches the bug: it fails with the exact missing-AKI error against a pre-fix build. Skips on windows and on LibreSSL pythons (Apple's system python ignores `SSL_CERT_FILE`, now documented in the proxy guide along with the no-proxy-credentials design).

## 1password: CLI batch reads fail on windows

`op run -- env -0` depends on a unix `env` binary, which doesn't exist on windows unless git's usr/bin is on PATH. Batch reads now use `op inject` with a stdin template of `op://` refs instead: same single authenticated call (batching performance unchanged), no child binary at all, and `--no-masking` is no longer needed.

Template entries are joined with a random per-call sentinel string rather than `\0` because `op inject` strips control characters from its output (verified against the real CLI, and the fake-op test harness mimics this). Multiline secret values are covered by a new test. References also no longer appear in the child process env, only on the stdin pipe.

### DIFF
--- a/.bumpy/1password-windows-inject.md
+++ b/.bumpy/1password-windows-inject.md
@@ -1,0 +1,5 @@
+---
+"@varlock/1password-plugin": patch
+---
+
+CLI batch reads now use op inject instead of op run -- env -0, fixing failures on Windows where no unix env binary exists
--- a/.bumpy/proxy-cert-strict-verifiers.md
+++ b/.bumpy/proxy-cert-strict-verifiers.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+proxy: minted MITM certs now include subject/authority key identifiers so strict TLS verifiers (python 3.13+ urllib/httpx defaults) accept them; Proxy-Authorization from clients is stripped instead of forwarded upstream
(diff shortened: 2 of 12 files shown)


theoephraim:
Split into #937 (proxy certs) and #938 (1password windows fix).

PR #943, opened 2026-07-27
Reclaim orphaned cache locks via owner liveness

### LINKED ISSUE #941: [BUG]: `awsParam()`/`awsSecret()` cache lock silently masks "parameter not found" behind an unhelpful "Timed out waiting for cache lock" error
### varlock version

1.13.0

### Summary

Adding `cacheTtl` to `@initAws(...)` turns a fast, clear `awsParam(): Parameter "..." not found` error into a slow (multiple minutes), confusing `Timed out waiting for cache lock at ~/.varlock/cache/varlock-default.json.keylocks/<hash>.lock` error, with no indication of the actual underlying problem (a missing SSM parameter).

### Repro

`.env.schema`:

```env
# @plugin(@varlock/aws-secrets-plugin)
# @initAws(region=$AWS_REGION, profile=$AWS_PROFILE, cacheTtl="1h")
# ---

ADMIN_ALARM_EMAIL=awsParam("/test/admin-alarm-email")
BASE_DOMAIN_NAME=awsParam("/test/base-domain-name")
```

Neither `/test/admin-alarm-email` nor `/test/base-domain-name` exist in SSM Parameter Store.

With `cacheTtl="1h"` set, `pnpm varlock load` hangs for well over a minute and then produces:

```
🚨🚨🚨  Configuration is currently invalid  🚨🚨🚨

⛔ ADMIN_ALARM_EMAIL*
   └ undefined
   - error resolving value: Error: Timed out waiting for cache lock at /Users/me/.varlock/cache/varlock-default.json.keylocks/5f2637b73df6c239fd1bb9a9c6de2b45890fd2626696f12453dfb9f0a35eeb6c.lock
⛔ BASE_DOMAIN_NAME*
   └ undefined
   - error resolving value: Error: Timed out waiting for cache lock at /Users/me/.varlock/cache/varlock-default.json.keylocks/9011085b51364bf54416c9dd387d0c2b476eb978676980a63b10a779c43f7420.lock

💥 Resolved config/env did not pass validation 💥
```

Removing `cacheTtl="1h"` (everything else identical) instead produces the correct error, immediately:

```
⛔ ADMIN_ALARM_EMAIL*
   └ undefined
   - awsParam(): Parameter "/test/admin-alarm-email" not found
     Verify the parameter exists in AWS Systems Manager Parameter Store
     AWS Console: https://console.aws.amazon.com/systems-manager/parameters?region=us-gov-west-1
⛔ BASE_DOMAIN_NAME*
   └ undefined
   - awsParam(): Parameter "/test/base-domain-name" not found
     Verify the parameter exists in AWS Systems Manager Parameter Store
     AWS Console: https://console.aws.amazon.com/systems-manager/parameters?region=us-gov-west-1
```

### Root cause (from reading `@varlock/aws-secrets-plugin`'s and `varlock`'s bundled source)

When `cacheTtl` is set, `awsParam()`'s resolver (`dist/plugin.cjs`) routes through a per-key file lock before calling the AWS SDK:

```js
// plugin.cjs
if (selectedInstance.cacheTtl !== void 0 && pluginCache) {
  const cacheKey = `awsParam:${instanceId}:${selectedInstance.cacheKeyIdentity}:${finalParameterName}`;
  const fullParameterValue = await pluginCache.getOrSet(
    cacheKey,
    selectedInstance.cacheTtl,
    async () => await selectedInstance.getParameter(finalParameterName)
  );
  ...
}
```

`CacheStore.getOrSet()` (in `varlock`'s `dist/chunk-PJQPUY5I.js`) acquires a directory-based lock keyed by a SHA-256 hash of the cache key before running the producer (i.e. the actual `getParameter()` call that would throw the clear "not found" error):

```js
var KEY_LOCK_OPTS = { waitMs: 50, timeoutMs: 5 * 6e4, staleMs: 10 * 6e4 }; // 5 min timeout, 10 min staleness

async getOrSet(cacheKey, ttlMs, producer) {
  ...
  const keyHash = createHash("sha256").update(cacheKey).digest("hex");
  const lockPath = path.join(`${this.filePath}.keylocks`, `${keyHash}.lock`);
  return await withDirLock(lockPath, KEY_LOCK_OPTS, async () => {
    ...
    const value = await producer(); // <-- this is where "Parameter not found" would normally throw
    ...
  });
}
```

`withDirLock()` acquires the lock via `fs.mkdirSync(lockPath)`. If that fails with `EEXIST` (lock dir already exists), it checks the lock directory's `mtime` against `staleMs` (10 minutes) to decide whether to reclaim it as abandoned; otherwise it polls every `waitMs` (50ms) until `timeoutMs` (5 minutes) elapses, then throws the generic `Timed out waiting for cache lock at ...` error — **without ever calling `producer()`**, so the real `ParameterNotFound` error is never reached or surfaced:

```js
async function withDirLock(lockPath, opts, fn) {
  const deadline = Date.now() + opts.timeoutMs;
  while (true) {
    try {
      fs.mkdirSync(lockPath, { mode: 448 });
      break;
    } catch (err) {
      ...
      if (err?.code !== "EEXIST") throw err;
      try {
        const stat = fs.statSync(lockPath);
        if (Date.now() - stat.mtimeMs > opts.staleMs) {
          // reclaim stale lock
          ...
          continue;
        }
      } catch {}
      if (Date.now() >= deadline) {
        throw new Error(`Timed out waiting for cache lock at ${lockPath}`);
      }
      await sleep(opts.waitMs);
    }
  }
  ...
  try {
    return await fn();
  } finally {
    clearInterval(touchTimer);
    fs.rmSync(lockPath, { recursive: true, force: true }); // only runs if the process reaches here
  }
}
```

The lock is only released in the `finally` block, which requires the process holding it to run to completion (including on error). **If a process holding the lock is interrupted before `finally` runs (Ctrl+C, SIGKILL, crashed terminal, killed shell, etc.), the lock directory is orphaned on disk** and isn't reclaimed until its `mtime` is more than `staleMs` (10 minutes) old. Any subsequent run that tries to resolve the same `awsParam()`/`awsSecret()` call in that window has no way to acquire the lock, so it just polls for the full `timeoutMs` (5 minutes) and throws the unrelated "Timed out waiting for cache lock" error — even though no other process is actually using it anymore.

**This is exactly what happened in my repro.** After an earlier `varlock load` attempt was interrupted while resolving these same two parameters, I found the orphaned lock directories still present on disk, with hashes matching the ones named in the error output:

```
$ ls ~/.varlock/cache/varlock-default.json.keylocks/
5f2637b73df6c239fd1bb9a9c6de2b45890fd2626696f12453dfb9f0a35eeb6c.lock
9011085b51364bf54416c9dd387d0c2b476eb978676980a63b10a779c43f7420.lock
```

Because the module resolves each `awsParam()` call independently and per-key, and because a missing/misconfigured parameter is a very plausible reason for a run to be slow/hang and get interrupted by an impatient user, this is an easy trap: with `cacheTtl` enabled, a single interrupted run leaves a lock that turns every subsequent `varlock load` attempt (while `cacheTtl` stays enabled) into a multi-minute, low-information failure, until ~10 minutes after the lock was orphaned, when it's reclaimed automatically. Disabling `cacheTtl` (as opposed to retrying with it still on) sidesteps the lock entirely and fails fast with the real error instead — which is why removing it, rather than waiting, is what actually resolved it here.

### What is expected?

Clear error message

Unclear error message

### System Info

```shell
System:
    OS: macOS 26.5.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 824.42 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @census/oxlint: workspace:^ => 0.1.0 
    @census/tsconfig: workspace:* => 0.1.0 
    @types/node: catalog: => 24.13.3 
    @typescript/native-preview: catalog: => 7.0.0-dev.20260331.1 
    @varlock/aws-secrets-plugin: catalog: => 1.2.2 
    aws-cdk: catalog: => 2.1132.1 
    esbuild: catalog: => 0.28.1 
    husky: catalog: => 9.1.7 
    oxfmt: catalog: => 0.57.0 
    oxlint: catalog: => 1.75.0 
    oxlint-tsgolint: catalog: => 7.0.2001 
    tsx: catalog: => 4.23.1 
    turbo: catalog: => 2.10.6 
    typescript: catalog: => 7.0.2 
    varlock: catalog: => 1.13.0 
    vite: catalog: => 6.4.3 
    vitest: catalog: => 4.1.10
```

### Suggested fixes

1. **Make the lock self-healing based on liveness, not just mtime.** Write the owning process's PID (and/or hostname) into the lock directory when it's created, and treat the lock as stale immediately if that PID is no longer alive (e.g. `process.kill(pid, 0)` throwing `ESRCH`), instead of waiting for the fixed 10-minute `staleMs` window. This is the standard approach most file-lock libraries use and would make orphaned locks self-heal almost immediately instead of blocking every subsequent run for up to 5 minutes.
2. **Register cleanup on interrupt signals.** `withDirLock`'s lock release currently only happens via the `try/finally` around `fn()`. Adding a `SIGINT`/`SIGTERM` handler (the codebase already does this elsewhere, e.g. `chunk-4QKDY7CZ.js`/`chunk-BOO3FAQT.js`, for other cleanup) that best-effort removes any locks the current process holds before exiting would prevent most orphaned locks from being created in the first place.
3. **Improve the timeout error message itself.** Even as a fallback, `Timed out waiting for cache lock at <path>` gives no next steps. Consider adding a tip along the lines of "If no other varlock process is running, this lock may be orphaned from an interrupted run — try deleting `<path>`" so users aren't left guessing.
4. **Reconsider the default `timeoutMs` (5 minutes) for `KEY_LOCK_OPTS`.** For a CLI dev-loop tool, waiting 5 minutes before surfacing *any* error is a long time, especially compared to the sub-second failure when caching is disabled.

### Workaround

Delete the stale lock directory (or the whole cache dir) and re-run:

```bash
rm -rf ~/.varlock/cache/varlock-default.json.keylocks
```

Or wait out the 10-minute `staleMs` window for the specific orphaned lock(s) to be reclaimed automatically.

theoephraim:
Fixes #941.

## The bug

A cache key lock left behind by an interrupted run (Ctrl+C at a biometric or password prompt, a killed shell) blocked every later run resolving the same key for the full 5 minute deadline, then failed with a generic `Timed out waiting for cache lock` that replaced the real error. In the reported case that hid an `awsParam(): Parameter "..." not found`.

This is in varlock core (`CacheStore.getOrSet`), not the AWS plugin, so it affects the 11 plugins that use `cacheTtl` plus the core `cache()` resolver.

## The fix

Lock dirs now record their owner (`pid`, `host`, `token`). A lock whose owner process is gone is reclaimed immediately, so the underlying error surfaces right away.

A **live** holder is still waited on, with no shortened timeout. That matters: the lock exists so that parallel task runners (turbo, nx) make one API call and show one biometric prompt per value instead of N. Shortening the deadline would have reintroduced exactly that. Timing out is now only a backstop for a holder that is alive but wedged, and it falls back to running the producer rather than failing the load.

Locks with no owner metadata (existing orphans, older versions) are reclaimed after a short grace window.

## Also fixed

- **Exclusion bug:** release now verifies the lock's token is still ours. Previously a holder that was correctly judged stale and stolen from would delete the *new* owner's lock on its way out, letting two processes run the same producer.
- **Re-entrancy:** a nested call on the same key proceeds instead of self-deadlocking. Tracked via `AsyncLocalStorage`, so concurrent siblings in the same process still dedupe (a process-global set conflates the two, which the existing dedupe test caught).
- **Signal cleanup:** held locks are released on SIGINT/SIGTERM/SIGHUP/SIGQUIT, using `once` + re-raise so `varlock run`'s signal forwarding and exit codes are unaffected.
- **`varlock cache clear` / `--clear-cache` now remove lock dirs**, and no longer skip when the cache file does not exist yet. Previously the obvious remedy silently did nothing.
- **Native encrypt is no longer `spawnSync`.** It ran inside the key lock, blocking concurrent item resolution and freezing the liveness heartbeat, which would have made a busy holder look dead.
- Lock timeout message names the remedy.

## Testing

`cache-lock.test.ts` covers liveness stealing, off-host locks, ownership tokens, fail-open, and re-entrancy. `cache-lock-crossproc.test.ts` spawns real child processes: one is SIGKILLed mid-lock to reproduce the actual bug, one verifies a live holder is not stolen from, and one verifies SIGINT both releases the lock and still terminates the process.

Verified the tests fail without the fix: disabling the liveness check fails exactly 3 tests and nothing else; disabling the re-entrancy short-circuit fails exactly 1.

## Docs

Added a "Concurrent runs" section and a troubleshooting entry to the caching guide.

### DIFF
--- a/.bumpy/cache-lock-liveness.md
+++ b/.bumpy/cache-lock-liveness.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+Cache locks left behind by an interrupted run are now reclaimed immediately instead of stalling later runs for several minutes and hiding the real error. `varlock cache clear` also clears locks.
--- a/packages/varlock-website/src/content/docs/guides/caching.mdx
+++ b/packages/varlock-website/src/content/docs/guides/caching.mdx
@@ -68,6 +68,14 @@ A few things to keep in mind:
 - The disk cache is per-OS-user and shared across all projects. This is intentional - projects often share config - but it means any env file you load can read and write it, so treat untrusted repos accordingly.
 - With the file-based encryption fallback, the decryption key sits on the same disk as the cache, so encryption is obfuscation-only (see warning above).
 
+## Concurrent runs
+
+When several varlock processes start at once (a turbo or nx pipeline, or a monorepo dev script), they coordinate through the disk cache so a value is only fetched once. The first process to need a given value fetches it while the others wait, then read the result from the cache. This is what keeps ten parallel tasks from triggering ten API calls, or ten biometric prompts, for the same secret.
+
+Coordination uses per-value lock directories under `~/.config/varlock/cache/`. If a process is killed while holding one (Ctrl+C at a password or biometric prompt, for example), the leftover lock is detected and reclaimed by the next run, so this is normally invisible.
+
+A lock can outlive its owner in one case: a cache directory shared between machines through a synced home directory, where varlock cannot tell whether the owning process is still running. `varlock cache clear --yes` removes lock directories along with entries.
+
 ## Per-invocation CLI controls
 
 [`varlock load`](/reference/cli-commands/#load), [`varlock run`](/reference/cli-commands/#run), and [`varlock printenv`](/reference/cli-commands/#printenv) support:
@@ -98,6 +106,16 @@ Notes:
 
 - If you used a custom `key`, that key may intentionally pin the cache entry.
 - Run with `--clear-cache` or clear entries via `varlock cache clear`.
+
+### "Timed out waiting for cache lock"
+
+A lock directory is being held and could not be reclaimed. Check whether another varlock process is genuinely running and still working. If none is, clear the locks:
+
+```bash
+varlock cache clear --yes
+```
+
+This is expected to be rare. Locks left behind by an interrupted run are reclaimed automatically on the next run.
 
 ### "`varlock cache` shows nothing, but caching seems active"
 
--- a/packages/varlock/src/cli/commands/cache.command.ts
+++ b/packages/varlock/src/cli/commands/cache.command.ts
@@ -191,12 +191,13 @@ const clearCommand = define({
     }
 
     try {
+      // a full clear also drops lock dirs. locks go first: clearing entries
+      // itself takes the file lock, and a wedged lock is often exactly why the
+      // user is running this command
+      if (!pluginName) store.clearLocks();
       const count = pluginName
         ? await store.clearByPrefix(`plugin:${pluginName}:`)
         : await store.clearAll();
-      // a full clear also drops lock dirs — clearing the cache is what users
-      // reach for when a lock is wedged, so it must actually unwedge it
-      if (!pluginName) store.clearLocks();
       console.log(`  Cleared ${count} ${target}`);
     } catch (err) {
       console.error(ansis.red(`Failed to clear ${target}: ${err instanceof Error ? err.message : err}`));
@@ -280,8 +281,8 @@ export const commandFn: TypedGunshiCommandFn<typeof commandSpec> = async () => {
         initialValue: false,
       });
       if (isCancel(confirmed) || !confirmed) continue;
-      const count = await store.clearAll();
       store.clearLocks();
+      const count = await store.clearAll();
       console.log(`  Cleared ${count} entries`);
       return;
     }
--- a/packages/varlock/src/cli/commands/cache.command.ts
+++ b/packages/varlock/src/cli/commands/cache.command.ts
@@ -194,6 +194,9 @@ const clearCommand = define({
       const count = pluginName
         ? await store.clearByPrefix(`plugin:${pluginName}:`)
         : await store.clearAll();
+      // a full clear also drops lock dirs — clearing the cache is what users
+      // reach for when a lock is wedged, so it must actually unwedge it
+      if (!pluginName) store.clearLocks();
       console.log(`  Cleared ${count} ${target}`);
     } catch (err) {
       console.error(ansis.red(`Failed to clear ${target}: ${err instanceof Error ? err.message : err}`));
@@ -278,6 +281,7 @@ export const commandFn: TypedGunshiCommandFn<typeof commandSpec> = async () => {
       });
       if (isCancel(confirmed) || !confirmed) continue;
       const count = await store.clearAll();
+      store.clearLocks();
       console.log(`  Cleared ${count} entries`);
       return;
     }
--- a/packages/varlock/src/env-graph/lib/loader.ts
+++ b/packages/varlock/src/env-graph/lib/loader.ts
@@ -49,7 +49,7 @@ export async function loadEnvGraph(opts?: {
       }
     }
     for (const store of diskStores) {
-      // locks are cleared unconditionally — a run interrupted before its first
+      // locks are cleared unconditionally: a run interrupted before its first
       // write leaves a lock dir behind with no cache file to gate on
       store.clearLocks();
       if (fs.existsSync(store.getFilePath())) await store.clearAll();
--- a/packages/varlock/src/env-graph/lib/loader.ts
+++ b/packages/varlock/src/env-graph/lib/loader.ts
@@ -49,6 +49,9 @@ export async function loadEnvGraph(opts?: {
       }
     }
     for (const store of diskStores) {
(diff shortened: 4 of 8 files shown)


theoephraim:
@pullfrog - review this

PR #959, opened 2026-07-29
fix(dashlane-plugin): don't hang forever on a locked vault (bound dcli calls, add onLocked option)

anuragbanerjee:
## Problem

`dashlane("dl://...")` resolves by spawning `dcli read <uri>` with piped stdio and no timeout. On a locked Dashlane vault, `dcli` prompts for the master password on a pipe that never answers, so the child never exits and any `varlock load` / `varlock run` whose schema contains a `dashlane()` entry hangs forever. This includes `@optional` entries the current command doesn't even need.

Repro:

```
dcli lock
varlock load --skip-cache   # schema with any dashlane() item: hangs forever
```

## Changes

1. **stdin closed on every `dcli` spawn** (`--version`, `sync`, `read`): interactive prompts hit EOF and fail immediately instead of blocking.
2. **Spawn timeout on every `dcli` call** as a backstop: default 30s, configurable via `@initDashlane(timeoutMs=...)`. A timeout maps to the existing "Dashlane vault appears locked or not synced" `ResolutionError` with an actionable tip.
3. **New `@initDashlane(onLocked=error|warn)` option**, default `error` (current behavior). With `warn`, the locked-vault error is thrown with warning severity, so `@optional` `dashlane()` items resolve empty with a surfaced warning and the load still passes; `@required` items still hard fail via `EmptyRequiredValueError`.

Behavior on a locked vault with `onLocked=warn`:

| Item | Before | After |
|---|---|---|
| `@optional` + `dashlane()` | hangs | resolves empty with a warning |
| `@required` + `dashlane()` | hangs | fails fast with a clear error |
| non-dashlane items | hangs | resolve normally |

With the default `onLocked=error`, everything fails fast with the locked-vault error instead of hanging.

## Small core change

`VarlockError` already supports warning severity, but `ConfigItem.resolve()` bailed on any resolution error before the required/optional check, so a warning-severity resolution error would have let `@required` items pass with only a warning. The bail now applies only to error-severity resolution errors: a warning-severity error leaves the value empty and normal required/optional handling takes over. Nothing in core currently throws warning-severity resolution errors, so this only affects plugins that opt in.

## Tests

New `locked vault handling` tests drive the existing fake `dcli` shim with new misbehavior modes: a blocked master-password prompt fails immediately (stdin closed), a hung call fails within `timeoutMs`, the `onLocked=warn` matrix above, and schema validation of the new options. Full dashlane plugin suite and varlock core suite pass, typecheck clean on both packages.

Docs updated in the plugin README and the website plugins page.

### DIFF
--- a/.bumpy/dashlane-locked-vault.md
+++ b/.bumpy/dashlane-locked-vault.md
@@ -1,0 +1,5 @@
+---
+"@varlock/dashlane-plugin": minor
+---
+
+dashlane() no longer hangs forever on a locked vault: dcli calls run with stdin closed and a timeout (default 30s). New @initDashlane options: onLocked=error|warn and timeoutMs
--- a/.bumpy/warning-resolution-errors.md
+++ b/.bumpy/warning-resolution-errors.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+warning-severity resolution errors no longer skip required/optional handling: the item resolves empty with the warning surfaced, and required items fail as empty
--- a/packages/plugins/dashlane/README.md
+++ b/packages/plugins/dashlane/README.md
@@ -178,6 +178,8 @@ Initialize a Dashlane plugin instance.
 - `id?: string` - Instance identifier for multiple instances (defaults to `_default`)
 - `autoSync?: boolean` - If `true`, runs `dcli sync` once before the first read (default `false`)
 - `lockOnExit?: boolean` - Lock the vault on process exit. Defaults to `true` in headless mode, `false` in interactive mode.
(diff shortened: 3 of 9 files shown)


theoephraim:
Thanks for this. Closing in favour of #989. I cherry picked some of this work, so you'll still get attribution.

Instead we add an `allowMissing` option to individual items and as an instance wide default. The idea being that if you wire up an item, it usually is expected to exist.

I'm not sold on coupling required/optional to any kind of resolution failure. Also not so sure about only warning on a locked vault. If you want to be able to resolve the schema in different situations (with the dashlane items or without) you should be able to use functions to do that, and explicitly attempt to load them or not.

PR #963, opened 2026-07-31
Add Vercel Sandbox guide

theoephraim:
Adds a Vercel Sandbox recipe to the sandbox guides, following the same shape as the E2B and Fly.io guides from #931: pass resolved values for trusted workloads, broker sandbox with the built-in tunnel for agentic ones, egress lockdown via the platform firewall, and a comparison note on Vercel's native firewall transforms.

The whole topology was validated live on real Vercel sandboxes, including wire substitution (hash oracle), response scrubbing, strict egress, and a live `networkPolicy` clamp. Vercel-specific facts baked into the recipe, all verified empirically:

- public sandbox URLs carry the WS tunnel as-is
- persistent sandboxes keep their domain, filesystem, and create-time env across stop/resume, so `getOrCreate` + `onResume` + `--persist-ca` makes broker wakes invisible to agents
- `npm i -g varlock` works on the default runtime
- default session timeout is 5 minutes, so the recipe always passes an explicit timeout

Docs only; no varlock code changes needed.


theoephraim:
Closing for now -- I think we actually want to hold off here and make something more like what we are planning for cloudflare.

PR #975, opened 2026-07-31
Add Box (ascii.dev) sandbox recipe

theoephraim:
Adds a cloud sandbox recipe for [Box](https://box.ascii.dev/) (ASCII's VM sandbox platform), plus the overview card and sidebar entry.

The recipe blesses the broker-box topology: one no-env box runs `varlock proxy start --expose`, agent boxes are created with `--no-env` and reach it via `proxy run --url`, over either the box's public IPv4 (after `ufw allow`) or their `host --public` HTTPS ingress (wss). All commands were verified live on Box (substitution, response scrubbing, strict egress deny, and the wss-over-ingress path).

Notable platform facts baked into the guide: default boxes receive real account credentials (model keys, GitHub token, Claude/Codex credential files) so agents should always be `--no-env`; boxes ship a default-deny inbound firewall; gated `host` routes use a redirect+cookie flow that WS upgrades cannot pass, so the tunnel route must be `--public` (varlock's data-plane token gates it). Box has no platform egress controls, so the guide carries the same credential-isolation-not-lockdown caveat as the smolvm recipe.

The lifecycle section is also live-verified: TTL auto-stop is absolute (default 1h, resets on resume) so an idle broker parks itself; there is no wake-on-connect; the IP changes on every resume (subdomain stays stable, and re-hosting the same port returns the same URL); `--persist-ca` survives stop/resume. One platform caveat documented: Box's docs say enabled systemd services restart automatically after resume, but in testing they did not, so the recipe has the orchestrator kick the service with one command after `box resume`.

Also fixes the stale docs build command in AGENTS.md (`--filter varlock-website` matches nothing; the package is `@varlock/website`).


theoephraim:
closing for now - will come back to it

PR #986, opened 2026-08-06
telemetry: honor the DO_NOT_TRACK env var

jdalton:
Thanks for varlock. The schema-driven approach to environment variables has been really useful.

`DO_NOT_TRACK` is an informal convention that a lot of command line tools follow _(Homebrew, Deno, Turborepo, Netlify CLI, Astro)_: the tool checks the `DO_NOT_TRACK` environment variable, and skips sending analytics when it is set. The point is that someone can set it once and opt out of every tool that honors it, instead of learning a different variable for each one. Homebrew, Deno, Turborepo, Netlify CLI and Astro all honor it today. varlock already has `VARLOCK_TELEMETRY_DISABLED`, a legacy `PH_OPT_OUT`, and config file flags, so this adds the standard variable next to them.

The change is in one function, `checkIsOptedOut()` in `packages/varlock/src/cli/helpers/telemetry.ts`. I pulled its environment variable check into a small exported helper so it could be unit tested without touching the real user config file, then added `DO_NOT_TRACK` there. The canonical value is `1`, but it is parsed with the same `TRUE_ENV_VAR_VALUES` list the existing variable uses, so `true` and `t` work too. Tests cover each branch: `DO_NOT_TRACK=1` opts out, `0` or unset or `false` does not, and the existing variables still behave the same. Docs are updated in the telemetry guide, the CLI reference, and the `varlock telemetry` command help.

📔 `PH_OPT_OUT` still matches only the exact string `true`, which is its existing behavior, and a test now asserts that so the legacy semantics do not drift.

### DIFF
--- a/packages/varlock-website/src/content/docs/guides/telemetry.mdx
+++ b/packages/varlock-website/src/content/docs/guides/telemetry.mdx
@@ -48,6 +48,12 @@ export VARLOCK_TELEMETRY_DISABLED=true
 
 This could be set in a specific terminal session, while running a specific command, in a Dockerfile, or in a CI/CD pipeline.
 
+varlock also honors `DO_NOT_TRACK`, a convention shared by many command line tools such as Homebrew, Deno and Turborepo. The convention is described at [donottrack.sh](https://donottrack.sh/). Setting it opts you out of every tool that supports it, so you do not need to remember a separate variable for each one:
+
+```bash
+export DO_NOT_TRACK=1
+```
+
 ### With a project config file
 
 You can also opt out at the project level by creating a `.varlock/config.json` file in your project root with the following content:
--- a/.bumpy/dnt-opt-out.md
+++ b/.bumpy/dnt-opt-out.md
@@ -1,0 +1,5 @@
+---
+'varlock': patch
+---
+
+Telemetry now honors the `DO_NOT_TRACK` environment variable, alongside the existing `VARLOCK_TELEMETRY_DISABLED` var and config file settings.
--- a/packages/varlock-website/src/content/docs/reference/cli/project.mdx
+++ b/packages/varlock-website/src/content/docs/reference/cli/project.mdx
@@ -180,7 +180,7 @@ varlock telemetry enable
 ```
 
 :::note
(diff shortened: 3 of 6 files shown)


packages/varlock-website/src/content/docs/guides/telemetry.mdx
```diff
@@ -48,6 +48,12 @@ export VARLOCK_TELEMETRY_DISABLED=true
 
 This could be set in a specific terminal session, while running a specific command, in a Dockerfile, or in a CI/CD pipeline.
 
+varlock also honors [`DO_NOT_TRACK`](https://consoledonottrack.com/), a convention shared by many command line tools. Setting it opts you out of every tool that supports it, so you do not need to remember a separate variable for each one:
```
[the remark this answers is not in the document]
jdalton: LOL,  kinda funny though. It could have been worse 🤏 🐕

theoephraim:
Seems reasonable. Thanks for the contribution! (And everything else)

packages/varlock-website/src/content/docs/guides/telemetry.mdx
```diff
@@ -48,6 +48,12 @@ export VARLOCK_TELEMETRY_DISABLED=true
 
 This could be set in a specific terminal session, while running a specific command, in a Dockerfile, or in a CI/CD pipeline.
 
+varlock also honors `DO_NOT_TRACK`, a convention shared by many command line tools such as Homebrew, Deno and Turborepo. The [original proposal](https://web.archive.org/web/20210708000000id_/https://consoledonottrack.com/) is linked through an archived copy, because the domain it used to live on no longer hosts it. Setting it opts you out of every tool that supports it, so you do not need to remember a separate variable for each one:
```
philmillman: I think we can just link to https://donottrack.sh/ instead of archive.org :)

jdalton: 🕺

philmillman (approved):
github seems to be having a major incident. this looks good to me, we'll wait to merge until the next release so the docs are in sync. thanks again @jdalton !

PR #999, opened 2026-08-12
fix(flatten): detect the workspace root in any language

### LINKED ISSUE #998: [BUG]: varlock flatten does not flatten .env.schema in uv (python) workspace package
### varlock version

1.16.1 (Standalone)

### Steps to reproduce

1. Create uv workspace with the following structure:
```shell
├── pyproject.toml
├── uv.lock
├── .env.schema
└── packages/
    └── foo/
        └── .env.schema   # contains: # @import(../../)
```
2. Run the following command from the workspace root:
```shell
cd packages/foo/ && varlock flatten
```

### What is expected?

```shell
Flattened 2 files into .env-flat/
  ..\..\.env.schema
  .env.schema

Add .env-flat/ to your .gitignore - it is a generated artifact.
```

```shell
No workspace root detected (no lockfile found) - imports reaching outside the current directory cannot be flattened
Flattened 1 file into .env-flat/
  .env.schema

⚠ @import(../../) in .env.schema points outside the workspace root and was left untouched - it must exist at that same path wherever the flattened output is used

Add .env-flat/ to your .gitignore - it is a generated artifact.
```

### System Info

```shell
System:
  OS: Windows 11 10.0.26200
  CPU: (32) x64 Intel(R) Core(TM) i9-14900K
  Memory: 94.91 GB / 127.71 GB
```

### Any additional comments?

Current workaround is to add an empty `package-lock.json` file in the workspace root. 

Not sure if `varlock flatten` is intended for use with non-JS workspaces.

philmillman:
Fixes #998

## The problem

`varlock flatten` determined the workspace root with `detectWorkspaceInfo()`, which only looks for **JS lockfiles**. In a uv (or Cargo, Composer, Gradle, .NET) monorepo there is no JS lockfile, so detection returned `undefined`, the root fell back to the package directory, and every `@import` reaching outside it was left untouched with a warning. The reporter's workaround was to drop an empty `package-lock.json` at the repo root.

## The fix

**New `detectWorkspaceRoot()` in `lib/workspace-utils.ts`**, separate from `detectWorkspaceInfo()`. The existing function stays as-is: "which package manager runs here?" genuinely wants the *nearest* lockfile, while "how far out does this repo extend?" does not.

- **Markers for every language varlock supports**: Python (`uv.lock`, `poetry.lock`, `pdm.lock`, `pixi.lock`, `Pipfile.lock`, `requirements.lock`), Rust (`Cargo.lock`), Go (`go.work` only, since `go.mod`/`go.sum` are per-module), PHP, Ruby, Elixir, JVM (`settings.gradle*`, `pom.xml`), .NET (`*.sln`/`*.slnx`, `global.json`, `Directory.*.props`), Bazel/Pants/Buck, plus the existing JS set. Per-package manifests (`pyproject.toml`, `Cargo.toml`, `package.json`) are deliberately excluded since every member has one.
- **Outermost match wins**, bounded by the VCS root. Ruby, PHP and .NET monorepos routinely carry a per-package lockfile, so nearest-match would pick the package itself.
- **`.git`/`.hg`/`.svn` is both the ceiling and the fallback root**, which covers repos with no lockfile at all (a Go repo with a `go.mod` per module).

**New `--workspace-root <path>` flag** for build contexts with neither a lockfile nor `.git`. Validated: must exist, and must contain the cwd (symlink-resolved before comparing).

**The command now prints the root it picked and how** (`Workspace root: ../.. (detected via uv.lock)`), which is what would have made the original report diagnosable from its output.

## Reviewer notes

- Detection does one `readdir` per ancestor directory rather than stat-ing each marker, which also gets `*.sln` matching for free.
- The walk stops at the user's home directory and at the filesystem root, so a stray lockfile above a repo can't be picked up.
- Only `flatten` uses the new function. `detectWorkspaceInfo()` and its callers are untouched.

## Verification

Ran end-to-end against the issue's exact layout plus Cargo, Ruby-with-nested-lockfile, a .NET solution, and a Go repo with only `.git`. The flattened uv package also loads standalone once `.env-flat/` is overlaid, resolving both its own key and the imported one.

36 new unit tests covering per-ecosystem markers, `.git` fallback (dir and file form, for worktrees), outermost-wins, the VCS ceiling, and the `--workspace-root` validation paths.

## Docs

Workspace root detection section with the full marker table in the flatten reference, plus notes in the Docker and monorepos guides.

### DIFF
--- a/.bumpy/flatten-workspace-root-detection.md
+++ b/.bumpy/flatten-workspace-root-detection.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+varlock flatten now detects the workspace root in any language (uv, poetry, cargo, go.work, composer, bundler, gradle, .NET, bazel), falls back to the git root, and accepts --workspace-root to set it explicitly
--- a/packages/varlock-website/src/content/docs/guides/monorepos.mdx
+++ b/packages/varlock-website/src/content/docs/guides/monorepos.mdx
@@ -131,6 +131,8 @@ Monorepos often combine Node apps with Python, Go, or other runtimes. Varlock's 
 
 Run the command from the service directory (or pass `--path` to its env folder). Each language can keep its own `.env.schema`; there is no requirement to share a single root file. For more detail, see [Other languages](/integrations/other-languages/).
 
+This works the same in a repo with no JavaScript in it at all. A uv, Cargo, Go, Composer, Bundler, Gradle, or .NET workspace is recognized as a monorepo, so [`varlock flatten`](/reference/cli/project/#flatten) can follow imports up to the repo root the same way it does in a JS workspace. See [workspace root detection](/reference/cli/project/#flatten) for the markers it looks for.
+
 ## Container builds
 
 Container images often copy a single app directory, not the whole monorepo. If your schema uses `@import()` to pull files from sibling packages or the repo root, those paths must exist in the build context, otherwise the load graph fails at build time.
--- a/packages/varlock-website/src/content/docs/integrations/docker.mdx
+++ b/packages/varlock-website/src/content/docs/integrations/docker.mdx
@@ -275,6 +275,7 @@ A few behaviors worth knowing (see the [command reference](/reference/cli/projec
 - `.env.local` / `.env.[env].local` files are **skipped by default**, so machine-local secrets don't end up in image layers. Use `--include-local` if you really want them.
 - Conditionally-enabled imports (`enabled=`) are copied too, with their conditions preserved, since production may enable imports your build machine doesn't.
 - `@plugin()` declarations in copied files get pinned to the installed version, so the standalone binary can [auto-install them](/guides/plugins/) in the container without the sibling package's `node_modules`.
+- The workspace root sets how far out imports can reach. It is detected from a lockfile or workspace config in any language (`uv.lock`, `Cargo.lock`, `go.work`, `composer.lock`, `pnpm-lock.yaml`, and so on), falling back to the `.git` root. Flatten prints the root it picked. If your build context has neither, pass [`--workspace-root <path>`](/reference/cli/project/#flatten).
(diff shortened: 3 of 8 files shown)


theoephraim:
Closing in favor of #1007, which fixes #998 by removing the workspace-root concept from `flatten` instead of extending detection.

The thing that decided it: with the resolution in `detectWorkspaceRoot`

```js
if (vcsRoot && outermostMarker?.rootPath !== vcsRoot.rootPath) return vcsRoot;
return outermostMarker ?? vcsRoot;
```

the loop breaks at the first `.git` walking up, so `outermostMarker` is always at or below `vcsRoot`. Whenever a `.git` exists the returned `rootPath` is the git root, unconditionally, and the 40-entry marker table can only change the `marker` label in the output. It changes the returned path only in a repo with no VCS directory at all. Since #998 is a git repo, the `.git` fallback alone is what fixes it.

That also makes the outermost-vs-nearest reasoning in the description moot in the Ruby/PHP/.NET cases it cites: git root wins over both.

Stepping back, `flatten` was using the root for a base directory to mirror imports under `.env-imports/` and for a refusal boundary. The base is answerable from the import graph itself, and the boundary was doing no real work: the user wrote those `@import` paths, varlock resolves them at runtime regardless, and refusing to copy them just produces output that breaks later. It was inconsistent too, refusing a file from a sibling repo while silently copying a gitignored one next door.

#1007 computes the base as the deepest directory containing the package and everything the import graph reaches, drops the boundary entirely, and replaces it with warnings for copied files that are outside the repo or gitignored. Same output wherever detection previously worked (22 existing tests pass unmodified), and no detection to keep current as ecosystems change.

The `--workspace-root` flag did not carry over. Nothing needs it once the base comes from the graph, but happy to add it back if a build context turns up that wants an explicit override.

PR #1018, opened 2026-08-18
feat(cloudflare): support wrangler preview command in varlock-wrangler

theoephraim:
Wrangler added a `preview` command for deploying branch previews (private beta). This teaches `varlock-wrangler` to handle it like the other deploy-style commands.

Requires wrangler >= 4.132.0, which shipped [cloudflare/workers-sdk#15256](https://github.com/cloudflare/workers-sdk/pull/15256) adding `--var` and `--secrets-file` to `wrangler preview`.

- Bare `varlock-wrangler preview` now routes through the deploy handler: non-sensitive values are passed as `--var` flags, sensitive values via `--secrets-file` (served over the existing FIFO), plus the `__VARLOCK_ENV` blob.
- Preview management subcommands (`preview delete`, `preview secret`, `preview base-config`, ...) are passed through to wrangler unchanged.
- `--keep-vars=false` is now only added for plain `deploy`; it was previously added for everything except `versions upload`, which would have leaked onto `preview` (which has no such flag).
- Docs: the deploy section of the Cloudflare integration page now covers `preview` and states the minimum wrangler version, plus where varlock resolves env files and how `--cwd` moves them (`--config` deliberately does not, so a single `.env.schema` still works with a wrangler config elsewhere).

### Deciding what counts as a preview deployment

Telling a preview *deployment* apart from a preview *management subcommand* needs two things wrangler owns and changes between releases: which options take a value (`preview --name delete` deploys a preview named "delete") and which subcommands exist. Options may also appear before the subcommand (`preview -c wrangler.jsonc delete`).

Rather than mirroring those tables, `varlock-wrangler` asks wrangler: appending `--help` makes yargs print the resolved command path as the first line of its output and exit without running anything. The probe is skipped when the args contain no positional at all (bare `preview`, `preview --json`), since a subcommand is always a bare word. If wrangler can't be reached or the output isn't a usage banner, the command passes through unchanged and wrangler reports its own error.

It also fixes two ways an invocation could silently skip injection, both pre-existing for `deploy` / `versions upload` / `dev` / `types` and inherited by `preview`:

- Wrangler accepts global flags before the command (`wrangler --cwd=/app deploy`), so the command is not always `argv[0]`. Routing now keys off the first bare word. Only the `--flag=value` form can precede the command (wrangler rejects `--cwd /app deploy` with `Unknown argument: /app`), so that word is always the command, never an earlier flag's value.
- `wrangler --cwd` runs as if started in another directory, but varlock resolved its graph from the caller's directory, so an enhanced command could deploy one project's worker with another project's values. The flag is now passed through to the varlock load. (The dev watcher needed no separate fix: its watch paths come from the graph's own `basePath`.)
- Wrangler reads anything after `--` as a positional, so flags appended there were silently dropped rather than parsed. They are now inserted before the terminator. Verified: `--secrets-file /nope.json` before `--` errors with `Could not read file`, after `--` it is never read.

That choice already paid off: between 4.93.0 and 4.132.0 wrangler dropped the `preview settings` subcommand and replaced it with `preview base-config`, and renamed `--ignore-defaults` to `--ignore-base-config`. A hardcoded list would have been wrong on both.


tnkuehne:
Heyo @theoephraim, https://github.com/cloudflare/workers-sdk/pull/15256 was merged. Can we get this merged?

theoephraim:
on it! should be out today

tnkuehne:
> on it! should be out today

Thanks for finishing and merging. Could you release the Cloudflare npm package

theoephraim:
@tnkuehne - sorry for the delay! Bunch of stuff going on over here... Working on getting this out today!

PR #1027, opened 2026-08-21
Add docs prose rules to AGENTS.md and clean up agent skills and apply to docs

philmillman:
## What changed

- **Docs style rules moved to a nested instructions file.** `packages/varlock-website/AGENTS.md` (new, with a `CLAUDE.md` shim) holds the docs and skill prose rules: the existing ones (no em dashes, no filler, prose only) plus eight new ones: no dismissive words (`just`, `simply`, ...), no `e.g.`/`i.e.`/trailing `etc.`, `must` vs `we recommend` instead of `should`/`usually`, check deep links before renaming headings, update `sidebar.ts` + redirects + cross-links when adding or moving pages, no hard-coded versions or duplicated lists (blog posts exempt), realistic names in code examples, delete stale content rather than keep it. The rules also cover user-facing READMEs (root and published packages). Root `AGENTS.md` keeps a short repo-wide summary and points at the nested file, so the full list is only loaded when working on docs or skills.
- **skills/varlock/SKILL.md**: applied those rules. Removed 36 em dashes, Latin abbreviations, and hedges; replaced the plugin and data-type lists with links to the reference pages; moved Setup up; dropped the "install this skill" step (covered by the installation and AI tools guides); install instructions now say a regular dependency (not a dev dependency); header now points at the Docs MCP and notes that any varlock.dev page can be fetched as markdown with `Accept: text/markdown`.
- **varlock-docs-search / varlock-agent-readiness skills**: added frontmatter (`name`, `description`) and rewrote the steps so they are actionable: markdown fetch and `llms.txt` before the MCP, the exact MCP tool name and what it returns, and a concrete digest verification procedure.
- **Docs site sweep (84 pages)**: applied the prose rules to every page under `src/content/docs/`: `e.g.`/`i.e.`/trailing `etc.` rewritten or replaced with links to the reference pages, dismissive words removed, `should`/`usually`/`typically` replaced with `must`, `we recommend`, or the actual condition, filler removed, hard-coded tool versions dropped from prose (genuine compatibility constraints kept). Code fences, inline code, frontmatter, and component markup are untouched. The code-examples rule is scoped to realistic names only (no runnable-as-shown requirement); placeholder names inside examples (`foo`/`bar`, `MY_VAR`, `KEY1`, ...) were replaced in a separate commit. Also fixed a stale claim in `sandboxes/minimal.mdx` (off-loopback bridging exists via `proxy start --expose`) and a typo in `reference/item-decorators.mdx`.
- **gen-agent-skills.ts**: reads `name`/`description` from each SKILL.md frontmatter (via `Bun.YAML`) instead of hard-coded strings, and throws if the frontmatter name does not match the directory, so `index.json` cannot drift from the files.

## Why

Came out of evaluating atinux/docster for the docs site. The skill itself is not a fit (it targets Comark/Nuxt and would misclassify a Starlight site), but a handful of its rules were worth adopting, and auditing our own skills against them turned up the violations fixed here.

## Notes for review

- Claims the sweep could not verify from source were softened rather than asserted; see the PR comment for the list worth a human look.
- The two remaining em dashes in the skill were inside a code fence's comments and were changed to colons.
- Website build passes and regenerates `index.json` with matching digests.


philmillman:
Claims the prose sweep could not verify from source. Each was reworded conservatively rather than asserted; worth a maintainer's eye:

- `guides/secrets.mdx` (build-output scanning): now says "most of our drop-in integrations do this automatically". Which integrations actually run a leak scan on build output?
- `guides/plugins.mdx` (caching): now says "via a `cacheTtl` init option, in most plugins". Confirm the option name and which plugins support it.
- `guides/import.mdx:21`: still says JSON/YAML/JS/TS import formats are planned "in the future". Still true?
- `integrations/docker.mdx` (Node section): keeps the "CLI is a few tens of MiB" memory claim.
- `integrations/cloudflare.mdx:398`: "a public API to build `__VARLOCK_ENV` bindings for non-wrangler deploy tools is planned but not available yet".
- `integrations/sveltekit.mdx:61`: "`@cloudflare/vite-plugin` doesn't currently support SvelteKit" (workers-sdk#8922). May be resolved upstream.
- `sandboxes/smolvm.mdx:53`: pinned to smolvm 1.7.1 and a fix merged 2026-07-30; left because it documents a real compatibility constraint, but it will need rewriting once a release ships.
- `guides/ai-tools/opencode.mdx`: a code fence uses `--model claude-3-5-sonnet`, an old model name. Inside a fence, so untouched.

Verified and fixed in this PR: `sandboxes/minimal.mdx` claimed off-loopback bridging did not exist; `proxy start --expose` implements it (`proxy.command.ts`). Verified and left as a recommendation: `@currentEnv` has no schema-only enforcement in source, so `reference/root-decorators.mdx` says "we recommend" rather than "must".

packages/varlock-website/src/content/docs/env-spec/reference.mdx
```diff
@@ -46,21 +46,21 @@ Comments in env-spec (like traditional .env files) start with a `#`. Unlike trad
 ```env-spec
    # ❌ leading space makes this invalid
 # this is a regular comment line
-# @dec2 @dec2=foo # this is a decorator comment line
-FOO=val # this is a post-value comment
-BAR=val # @dec # post-value comments may also contain decorators
+# @dec1 @dec2=val # this is a decorator comment line
```
theoephraim, line 49: non-meaningful placeholders kind of make sense in a case like this - like it sort of lowers mental load?

packages/varlock-website/src/content/docs/getting-started/installation.mdx
```diff
@@ -109,7 +109,7 @@ cosign verify-blob checksums.txt \
 Both workflow identities are accepted because binaries are normally cut by `release.yaml` alongside the npm publish, and re-cut by `binary-release.yaml` when an already-published version needs new archives.
 
 :::note
-Signing was added after varlock 1.14.0, so older releases have `checksums.txt` but no bundle. If a release has no `checksums.txt.cosign.bundle` asset, it predates signing and only the checksum step applies.
+Older releases have `checksums.txt` but no bundle. If a release has no `checksums.txt.cosign.bundle` asset, it predates signing and only the checksum step applies.
```
theoephraim, line 112: ok to leave version numbers when we are talking about when something is released or a floor

packages/varlock-website/src/content/docs/guides/proxy/rules.mdx
```diff
@@ -58,17 +58,17 @@ This injects `STRIPE_SECRET_KEY` across `api.stripe.com` and blocks refunds and
...
+- `strict`: add `@proxyConfig={egress="strict"}` to the header so only requests that match an allow (`@proxy`) rule are allowed; everything else is blocked, **including a request to a host that has `@proxy` rules but none matching this path/method**. This is the recommended posture once your rules are dialed in, since it prevents the agent from reaching arbitrary hosts (or arbitrary endpoints on a routed host).
 
-A matching `block` rule **always wins over an allow rule**, in either egress mode — so a `block=true` rule denies a request even if a broader `@proxy` allow rule also matches it. (To allow only a subset of a host and deny the rest, use `strict` egress with a specific allow rule, rather than a broad `block` with a narrow allow.)
+A matching `block` rule **always wins over an allow rule**, in either egress mode, so a `block=true` rule denies a request even if a broader `@proxy` allow rule also matches it. (To allow only a subset of a host and deny the rest, use `strict` egress with a specific allow rule, rather than a broad `block` with a narrow allow.)
```
theoephraim, line 63: probably ok to still use regular dashes?

packages/varlock-website/src/content/docs/guides/proxy/running.mdx
```diff
@@ -49,7 +49,7 @@ When the guest's only egress is an explicit HTTP proxy (a corporate proxy, or a
 
 Every `varlock proxy` command operates on a **session**: one running proxy with its own short id (printed by `proxy start`, listed by `proxy status`). You target a session in one of two ways:
 
-- **By id:** pass `--session <id>` to any command (`proxy env --session abc12`, `proxy stop --session abc12`, etc.).
+- **By id:** pass `--session <id>` to any command (for example `proxy env --session abc12` or `proxy stop --session abc12`).
```
theoephraim, line 52: dont have strong opinions on swapping ".e.g." to "for example", but probably lots of cases like this where neither is needed, and less text is probably better?

packages/varlock-website/src/content/docs/getting-started/installation.mdx
```diff
@@ -126,7 +126,7 @@ npx skills add dmno-dev/varlock
   </TabItem>
   <TabItem label="GitHub CLI">
 
-Requires GitHub CLI v2.90+.
+Requires a GitHub CLI version that includes the `gh skill` command.
```
theoephraim, line 129: same on this one

Issue #428, opened 2026-03-17
[BUG]: environment flag "..." must be defined within this schema when importing


bestickley:
### varlock version

0.4.2

### Steps to reproduce

.env.schema
```
# @currentEnv=$DEPLOY_ENV
# @defaultSensitive=false
# @generateTypes(lang=ts, path=./src/env.d.ts, executeWhenImported=true)
# @import(
#   ../../../.env.schema,
#   AWS_ACCOUNT,
#   AWS_PROFILE,
#   AWS_REGION,
#   DEPLOY_ENV,
#   SHARED_INFRA_NAMESPACE,
# )
# ---
AUTH_TABLE_NAME=awsParam("/$SHARED_INFRA_NAMESPACE/$DEPLOY_ENV/auth-table-name")
CHAT_URL="http://localhost:8000/api/v1/chat"
```

### What is expected?

success

### What is actually happening?

error:
```
pnpm varlock load
🚨 Error encountered while loading .env.schema

environment flag "DEPLOY_ENV" must be defined within this schema

Stack trace:
Error: environment flag "DEPLOY_ENV" must be defined within this schema
    at DotEnvFileDataSource.finishInit (file:///Users/stickb/Code/census/llm-research/node_modules/.pnpm/varlock@0.4.2/node_modules/varlock/dist/chunk-PCRIVT4T.js:10166:30)
    at async DirectoryDataSource.addChild (file:///Users/stickb/Code/census/llm-research/node_modules/.pnpm/varlock@0.4.2/node_modules/varlock/dist/chunk-PCRIVT4T.js:10046:5)
    at async DirectoryDataSource.addAutoLoadedFile (file:///Users/stickb/Code/census/llm-research/node_modules/.pnpm/varlock@0.4.2/node_modules/varlock/dist/chunk-PCRIVT4T.js:10490:5)
    at async DirectoryDataSource._finishInit (file:///Users/stickb/Code/census/llm-research/node_modules/.pnpm/varlock@0.4.2/node_modules/varlock/dist/chunk-PCRIVT4T.js:10495:5)
    at async DirectoryDataSource.finishInit (file:///Users/stickb/Code/census/llm-research/node_modules/.pnpm/varlock@0.4.2/node_modules/varlock/dist/chunk-PCRIVT4T.js:10107:5)
[...]

PR #1036, opened 2026-08-27
fix: an imported @currentEnv must win over the --env fallback

chriscors:
### The bug

A directory whose schema gets `@currentEnv` through `@import()` never loads its own `.env.[env]` files when a fallback env is supplied.

Minimal repro — same directory, same files, the only difference is the flag:

```console
$ varlock load
BETTER_AUTH_URL = "https://app-level-stg.example.com"     # app's .env.stg wins

$ varlock load --env production
BETTER_AUTH_URL = "https://root-level-stg.example.com"    # app's .env.stg silently skipped
```

with:

```
apps/dashboard/.env.schema   # @import(../../)
apps/dashboard/.env.stg      BETTER_AUTH_URL=https://app-level-stg.example.com
.env.schema                  # @currentEnv=$APP_ENV  (+ APP_ENV declared here)
.env.stg                     BETTER_AUTH_URL=https://root-level-stg.example.com
```

`APP_ENV=stg` throughout. Note the root's `.env.stg` **does** load — so `@currentEnv` resolves fine — while the app's own is skipped in favour of a `.env.production` that doesn't exist. No error, and nothing in the loaded-file list hints the app file was considered:

```
- Environments: .env.schema, ../../.env.stg, ../../.env.schema
```

### Why

`DirectoryDataSource._finishInit()` resolves `currentEnv` **before** imports are processed, which is too early for an imported `@currentEnv`, so it falls through to `graph.envFlagFallback`. That value is truthy, so the post-import re-check — guarded on `!currentEnv` — never runs, and the directory stays pinned to the fallback for the rest of the load.

There's no fallback in the CLI's default path, so this only surfaces when one is set. The Next.js integration always sets one: `next-env-compat.ts` passes `--env development`/`production` from `next dev`/`next build` to match `@next/env`. Its comment says the user should be able to *"ignore it by setting their own `@currentEnv`"* — which holds when the decorator is in the same schema (there's already a passing test for that), and doesn't when the schema imports it.

That makes the monorepo layout in [Monorepos → one schema per project](https://varlock.dev/guides/monorepos/) unusable for per-environment values under Next.js: each app's `.env.schema` is just `@import(../../)`, so no app can have its own `.env.[env]`. Worse, both staging and production deploys run `next build` → `--env production`, so they resolve identically.

### The fix

`_resolveCurrentEnv()` now reports whether the value came from the fallback. A fallback is treated as provisional:

- it is **not acted on before imports**, and
- the post-import re-check runs for it, not only when there was no value at all.

Deliberately *not acted on early* rather than loaded and later overridden: loading `.env.<fallback>` leaves its values in the graph even once the real env is known, so a key present only there would leak into the wrong environment.

A resolved `@currentEnv` is unchanged — still final, still loaded before imports, so import conditions can read those values.

### Tests

One test added next to the existing `fallback env value is ignored if currentEnv is present`, covering the same intent when `@currentEnv` arrives via an import. It fails on `main`:

```
AssertionError: ITEM1 value did not match: expected 'val-from-.env.staging' to deeply equal 'val-from-.env.dev'
```

Full suite green with the fix: **1797 passed, 1 skipped**. `typecheck` and `eslint` clean.

Also verified end to end against the real monorepo this came from — building the patched CLI and running the exact command the Next integration shells out to (`load --env production`) now resolves the app-level value.

### Related

#428 is a different symptom of `@currentEnv` + `@import` (env flag must be declared in the same schema when using a pick list); this path is separate and doesn't fix that one.

### DIFF
--- a/.bumpy/fix-imported-currentenv-vs-fallback.md
+++ b/.bumpy/fix-imported-currentenv-vs-fallback.md
@@ -1,0 +1,5 @@
+---
+varlock: patch
+---
+
+Fix `@currentEnv` from an import losing to the `--env` fallback when loading a directory's own `.env.[env]` files
--- a/packages/varlock/src/env-graph/lib/data-source.ts
+++ b/packages/varlock/src/env-graph/lib/data-source.ts
@@ -935,8 +935,15 @@ export class DirectoryDataSource extends EnvGraphDataSource {
     await this._initChild(child);
   }
 
-  /** resolve currentEnv from schema's envFlagKey or parent chain */
-  private async _resolveCurrentEnv(): Promise<string | undefined> {
+  /**
+   * resolve currentEnv from schema's envFlagKey or parent chain
+   *
+   * `fromFallback` distinguishes a value that came from the graph-level fallback
+   * (`--env`, which the Next.js integration sets from `next dev`/`next build`) from one
+   * resolved via an actual `@currentEnv`. The fallback is only a guess until imports
+   * have been processed, since an import may introduce a real `@currentEnv`.
+   */
+  private async _resolveCurrentEnv(): Promise<{ env: string | undefined, fromFallback: boolean }> {
     if (!this.graph) throw new Error('expected graph to be set');
 
     // First check if our schema has its own envFlagKey (for partial imports with their own currentEnv)
@@ -951,16 +958,18 @@ export class DirectoryDataSource extends EnvGraphDataSource {
           + `but "${envFlagKey}" is not included in the import list. `
           + `Add "${envFlagKey}" to the @import() arguments.`,
         ));
-        return undefined;
+        return { env: undefined, fromFallback: false };
       }
       const envFlagItem = this.graph.configSchema[envFlagKey];
       if (envFlagItem) {
         if (!envFlagItem.resolvedValue) await envFlagItem.earlyResolve();
-        return envFlagItem.resolvedValue?.toString();
+        return { env: envFlagItem.resolvedValue?.toString(), fromFallback: false };
       }
     }
     // Fall back to parent chain or fallback value
-    return (await this.resolveCurrentEnv())?.toString() || this.envFlagValue?.toString();
+    const fromEnvFlagItem = !!this.envFlagConfigItem;
+    const env = (await this.resolveCurrentEnv())?.toString() || this.envFlagValue?.toString();
+    return { env, fromFallback: !!env && !fromEnvFlagItem };
   }
 
   /** load env-specific files (.env.ENV and .env.ENV.local) for the given environment */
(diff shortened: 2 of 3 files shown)


theoephraim:
Thanks for this! Will have to dig in a bit. I know it was made considerably more complex by attempting to match next’s default behavior of setting env based on running build vs dev (if no currentEnv is set). I’d rather not do that, but we were trying to make it work as a drop in replacement.

chriscors:
Good catch — that's a real regression, and I've reverted the part of the change that caused it.

I verified it rather than taking it on faith: your case as a test **passes on `main` and failed on b012969**, with exactly the error you predicted:

```
"AUTH_MODE" was already resolved during early initialization
(used by @currentEnv, @import enabled, or @disable) and cannot be redefined by .env.staging
```

Deferring made `AUTH_MODE` early-resolve to `none`, so the conditional import was skipped and the later file then collided with that decision. It's now a permanent test (`fallback env values are available to import conditions`), so the deferral can't come back by accident.

**What changed in 0f25087:** `.env.<fallback>` loads before imports exactly as it did on `main`. The only difference from `main` is the post-import re-check, which now also runs when the value came from the fallback — not just when there was no value at all. If a real `@currentEnv` then resolves to a different env, that env's files load after the fallback's and take precedence.

**On your second required outcome** — "values from the fallback-specific files must not remain in the graph" — I have not done that, deliberately. Dropping them means removing a data source *and* the config items it already created, and I couldn't find a mechanism for that; inventing one felt like a much bigger change than this bug warrants, and easy to get subtly wrong. So a key present **only** in `.env.<fallback>` will still be set after the fallback is superseded. It's noted in a comment at the site.

Happy to take a swing at it if you'd point me at the right seam — or to leave it as a documented edge, since in the motivating case nothing loads early anyway (`.env.production` doesn't exist).

Full suite green: **1798 passed, 1 skipped**. Typecheck and lint clean.

chriscors:
Certainly! Noticed while trying to implement multi-env overrides across multiple next apps. Let me know if I can provide any add'l followup.

chriscors:
Confirmed as real — and it fails on `main` as well, so it isn't something this PR introduces.

I built your case as a test:

```ts
fallbackEnv: 'staging',
files: {
  '.env.schema': `# @import(./shared/)
                  # @import(./.env.azure, enabled=eq($AUTH_MODE, "azure"))
                  # ---
                  AUTH_MODE=none`,
  'shared/.env.schema': `# @currentEnv=$APP_ENV
                         # ---
                         APP_ENV=dev`,
  '.env.staging': 'AUTH_MODE=azure',
  '.env.azure':   'AZURE_ITEM=val-from-.env.azure',
},
expectNotInSchema: ['AZURE_ITEM'],
```

`AZURE_ITEM` is present both on `main` and on this branch. On `main` the directory never resolves the imported `@currentEnv` at all, so it stays on `staging` and the azure import is enabled by the fallback — the same end state, just self-consistently wrong rather than inconsistently.

**The two findings can't both be satisfied in one pass.** Your first one requires `.env.<fallback>` to load *before* imports, because import conditions read those values. This one requires import decisions to reflect an `@currentEnv` that only exists *after* imports are processed. That's circular: the env flag is defined by an import, and the imports are gated on values selected by the env flag.

Breaking it needs a design change, and I don't think it belongs in a bug-fix PR from a first-time contributor. The two shapes I can see:

1. **Env-flag pre-pass** — process only the imports that can define the env flag, resolve `@currentEnv`, then process the remaining (conditional) imports with the right env. Cheapest, but needs a rule for which imports are "env-flag-defining", and is still circular if such an import is itself conditional.
2. **Two-phase load with teardown** — load speculatively under the fallback and, if the resolved env differs, discard the directory's sources and reload. Correct by construction, but needs source/item removal, which is the same machinery your other comment's "must not remain in the graph" would need.

What this PR does is strictly narrower: the env flag now resolves correctly through an import, which is what makes per-app `.env.[env]` files work at all in a monorepo. It doesn't make the conditional-import case worse — that behaves identically to `main`.

So: happy to leave the PR scoped here, or to attempt (1) or (2) if you tell me which you'd prefer and whether the removal machinery is something you'd want to exist. I can also add the case above as a skipped/`todo` test documenting the limitation, if that's useful to you — say the word and I'll push it.

theoephraim:
Thanks again for digging into this. Your original commit (b012969) had the right semantics: `--env` should be a true last resort, and the review feedback that pushed you off it was from an automated reviewer asking for two mutually exclusive things. I've cherry-picked that commit as-is (your authorship) into #1050 so it lands cleanly on main. Closing this in favor of that one.

PR #1053, opened 2026-08-31
Stop deleting runtime env vars when booting from a build-baked snapshot

theoephraim:
Fixes a destructive silent failure reported from production: a Next.js standalone container passing env vars at boot (`docker run -e REDIS_URL=...`) had those vars **deleted from `process.env`** by varlock at startup, taking the service down.

## Root cause

In a container where the varlock CLI is unreachable, the standalone server falls back to the env blob baked into the bundled runtime at build time. `initVarlockEnv` treated that blob like a fresh resolution, so items that resolved to undefined at build time (`REDIS_URL=` in the schema, no value in the build env) triggered the stale-echo cleanup from #1038, deleting the runtime-provided value. That cleanup assumes a resolution happened in this process (a genuine ambient value would have acted as an override and resolved to it), which is false for a blob resolved on a build machine.

The deletion shipped only in `varlock@1.17.1`; 1.17.0 and earlier are unaffected.

## Behavior change

Scoped deliberately to the regression plus visibility, so no currently-booting deployment breaks:

- **Runtime-provided vars are never deleted** on a baked boot. This is the fix.
- **The baked snapshot stays authoritative for `ENV`**, as before: it was resolved at build time, and runtime values cannot be validated or coerced against it (the blob carries no schema info).
- **A runtime value that differs from the snapshot is logged as a warning** naming the keys, instead of being ignored in silence. The boot is never blocked, and absence is not a mismatch, so blob-only deploys (serverless, deliberate bake-into-image) are untouched.
- Values that cannot be verified (coerced values with no recorded raw pre-coercion form, e.g. sensitive `FLAG=YES` → `true`) are skipped rather than reported.

Enforcing the mismatch (failing the boot) was deliberately left out: silent-ignore of runtime env on a baked boot predates this bug and exists in every released version, and failing on it would take down deployments that work today (docker sets `HOSTNAME` in every container; promoted images and `fly secrets set` drift the same way). That belongs with `varlock freeze` (#1049), where baking becomes explicit and the contract can be enforced honestly.

## Mechanics

- The injection preludes (nextjs webpack + turbopack, vite `resolved-env` SSR entry) bake an `injectedAtBuild: true` flag INSIDE the serialized payload, before encryption where applicable. Provenance therefore travels with the blob to child/worker processes and through encryption round-trips, and can never outlive it: any fresh resolution produces an unflagged blob, so no marker clearing is needed anywhere.
- For the nextjs preludes a blob already present in the runtime env (`varlock run`) wins over the baked fallback, as before. The vite `resolved-env` payload stays authoritative by design: that artifact is frozen on purpose, and its static values are inlined at build time, so honoring a boot-time blob would override only the runtime-resolved subset.
- The `@next/env` compat does not trust a flagged blob as its cached env and attempts a real load instead, covering runtimes that load before the compat (e.g. via `instrumentation.js`).
- Fixes a pre-existing crash on the side: in `init-only` mode an encrypted ambient blob (`varlock run --inject blob` under `@encryptInjectedEnv`) reached `initVarlockEnv` as ciphertext and threw. The vite prelude now decrypts it, mirroring the nextjs init-server behavior.

## Testing

- Unit coverage for the no-delete fix, the mismatch warning, unverifiable coerced values, and unflagged (freshly resolved) blobs.
- Framework-test scenarios booting a real standalone build copied outside the project tree with a stripped environment (mimicking a container image), per bundler. Vite scenarios cover the mismatch warning, an ambient blob (baked payload still wins), and the `init-only` encrypted-blob decrypt.
- Green on Next 14/15/16 and the full vite and nuxt suites.

## Docs

The environments guide now clarifies that process-env overrides apply at resolution time; the Next.js, vite, nuxt, and Docker pages describe what a baked boot does with runtime env and how to make runtime values apply.

Follow-ups (separate PRs, discussed): land `varlock freeze` (#1049) as the single explicit pinning mechanism, then rework the integrations so baking is explicit or platform-automatic, and enforce the frozen contract there.


theoephraim:
Superseded by #1055, which is the regression fix alone (~250 lines vs 585).

The scope here drifted a long way past the reported bug: a mismatch warning, a `next-env-compat` cached-env change, an unrelated `init-only` vite decrypt fix, and docs describing a fail-closed design that the final commit removed. The framework tests on this branch still assert the removed throw (`conflicts with the build-time env snapshot`), so the branch no longer builds a coherent story.

#1055 keeps the one thing that was hard-won here: the `injectedAtBuild` flag lives inside the payload, not beside it. Everything else moves to `varlock freeze` (#1049) or its own PR.

PR #1058, opened 2026-09-01
Add remote HTTP approval provider

tnkuehne:
Hey together, 
thanks a lot for building Varlock. Makes my life a lot easier when working with secrets and agents. And also thanks a lot @theoephraim for quick reactions (https://github.com/cloudflare/workers-sdk/pull/15256). Hope it gets merged soon. 

But now to the roadblock I hit. Lately I am using remote dev machines, which means my normal secret approval flows with varlock dont work there. So i gave codex some homework to research and it found you guys (of course) already thought about the problem but the final implementation is not done yet.

So I gave it some thought and had codex work on it. Reviewed everything and tested it.

Would love to get your thoughts on it, if this is going in the right direction I am happy to give this some more work as this is currently one of my biggest pains. 

I did not yet add any tests, as I wanted to hear your thoughts. But will happily add some if this is going in the right direction.

Below some thought I wrote together with codex:

----

# Remote HTTP approval provider

Varlock already supports request-bound approvals, lifetimes, and grants. This change makes that approval mechanism usable when the operator is not at the broker terminal.

## What we added to Varlock

- An HTTP implementation of the existing `ApprovalProvider`.
- Broker authentication through a small `HttpApprovalAuth` interface, with Bearer authentication as the initial CLI implementation.
- Signed-decision verification through an `ApprovalDecisionVerifier` interface, with Ed25519 as the initial implementation.
- CLI and environment configuration for the service URL, Bearer token, and pinned public key.
- An `ApprovalPendingNotification` containing the approval URL and safe request metadata.
- An event channel over the existing authenticated tunnel so `proxy run --url` can display that notification remotely.
- Removal of approval-service credentials from the proxied child environment.

The existing terminal provider remains the default.

## HTTP approval flow

When an existing `require-approval` rule matches:

1. Varlock sends the existing request-bound `ApprovalRequest` to the configured service.
2. The service returns an operator-facing `approvalUrl` and broker-facing `statusUrl`.
3. Varlock displays the approval URL locally and forwards it to attached remote clients.
4. The broker polls the status URL.
5. An approval is accepted only if its nonce, expiry, lifetime, and signature are valid.
6. Varlock then applies its existing lifetime clamping and grant behavior.

## Security decisions

The approval URL may be visible to the agent, so possession of it cannot authorize a decision. Operator authentication belongs to the approval service.

The Bearer token and Ed25519 key have separate roles:

- The Bearer token authenticates broker-to-service requests.
- The approval service signs the exact decision with its Ed25519 private key.
- Varlock holds only the public key and verifies the decision locally.

The signature binds the request method, verified host, path, body hash, nonce, expiry, rule, grant scope, secret names, and selected lifetime. A decision therefore cannot be reused for a different request or widened beyond what was approved.

Errors, redirects, malformed responses, invalid proofs, nonce mismatches, and timeouts fail closed. HTTPS is required outside loopback, and the status URL must use the configured service origin.

## Remote notification

The existing tunnel continues carrying proxied traffic. We add a separate authenticated event connection that carries only pending-approval notifications.

The remote client can display the approval link but cannot submit a decision through the tunnel. Decisions still travel directly between the broker and approval service.

## What we reuse unchanged

- The existing `ApprovalRequest` and `ApprovalProvider` model.
- Existing `once`, `session`, and `duration` lifetimes.
- Existing grant keys, duration limits, and grant storage.
- Existing proxy rules and `require-approval` behavior.
- The existing authenticated WebSocket tunnel.

## Reference service

We also built a separate Cloudflare Worker to validate the protocol. It stores pending approvals in Durable Objects, authenticates the operator with WebAuthn, and signs approved decisions with Ed25519. It is one possible implementation of the generic HTTP approval contract.

### DIFF
--- a/.bumpy/http-approval-provider.md
+++ b/.bumpy/http-approval-provider.md
@@ -1,0 +1,5 @@
+---
+varlock: minor
+---
+
+Proxy: add a fail-closed HTTP approval provider with signed, request-bound decisions and forward pending approval links to remote `proxy run --url` clients over the authenticated tunnel.
--- a/packages/varlock-website/src/content/docs/guides/proxy/running.mdx
+++ b/packages/varlock-website/src/content/docs/guides/proxy/running.mdx
@@ -45,6 +45,58 @@ The workload does not have to be on the same machine. `proxy start --expose` mak
 
 When the guest's only egress is an explicit HTTP proxy (a corporate proxy, or a sandbox gateway like [Docker Sandboxes](/sandboxes/docker-sandboxes/)), `proxy run --url` dials the tunnel through it automatically: it honors `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` and `NO_PROXY` (a loopback or `NO_PROXY`-matched broker still dials direct). Only a plaintext `http://` `CONNECT` proxy is supported (an `https://` proxy, meaning TLS to the proxy itself, and SOCKS are not); a `wss://` broker still works through it, since the tunnel's own TLS runs end to end inside the `CONNECT`. That end-to-end TLS also means a proxy that terminates TLS only ever sees the encrypted tunnel.
 
+### HTTP approvals
+
+Use `--approval-url` on the machine that owns the proxy to replace its terminal prompt with an HTTP approval service:
+
+```bash
+VARLOCK_APPROVAL_TOKEN='broker-to-service-token' \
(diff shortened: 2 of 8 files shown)


theoephraim:
You are seeing into the future :)  This is exactly what we are building up towards.

First approvals on the device itself (especially on Mac with a fingerprint reader) - then cross device with a blind relay hosted on cloudflare as you were playing with. Plan is that this will be a hosted service we provide, along with some identity and device management. But could potentially make it open - so will think that through with this PR in mind.

PR #1068, opened 2026-09-04
Fix plugin cache policy handling, add producer-derived TTLs, declare generated grammar as a build output

theoephraim:
Three independent fixes, split out of #1044 so they can land without waiting on the AWS MFA feature. None of them are AWS-specific.

## Plugin caches ignored the `@cache` policy

Plugins install during `finishInit`, before the `@cache` root decorator is applied in `finishLoad`. A cache accessor bound to whatever store existed at install time therefore never saw the policy, so under `@cache=memory` or `@cache=disabled` every plugin's `cacheTtl` fell through to a no-op store. That affects all 11 plugins that cache, not just one.

The accessor now looks up the backing store per call, and the final store is re-propagated to plugins once the policy resolves.

Covered by a new test using a fixture plugin that resolves through `plugin.cache`: two items sharing a key must produce one value. Without the fix the producer runs twice.

## `getOrSet` can derive its TTL from the produced value

`getOrSet` took its TTL up front, but for anything whose lifetime the source decides (an STS session, an OAuth token, a lease) that is only known once the producer has run. Callers had to guess a lower bound and correct it afterwards, leaving a window where the entry outlived what it described.

The `ttl` argument now also accepts a callback receiving the produced value, called once and only when the producer actually produced. Returning zero or less skips the write. All existing callers pass a literal and are unaffected.

Note the guarantee is scoped: cross-process sharing covers persisted results only. A skipped or failed write leaves later processes to run the producer themselves, so a caller guarding single-use work needs a positive TTL. The docstring says so.

## Generated parser grammar was not a declared build output

`packages/env-spec-parser/src/grammar.js` is produced by peggy and gitignored, but the build task only declared `dist/**`. On a cache hit turbo restored `dist/` without it, so a fresh checkout that hit the cache had no grammar and anything importing the parser from source failed with `ERR_MODULE_NOT_FOUND`. This surfaced whenever a downstream test task cache-missed while the parser build cache-hit.

Also adds `grammar.peggy` to the build inputs. It sits at the package root, so the shared `src/**` glob never covered it and editing the grammar would not have invalidated the cache.

The eslint jsonc glob now matches package-level `turbo.json` files, not just the root one, so they can carry comments the way the root config does.


theoephraim:
Confirmed and fixed in 1609aa7c by adding `"env": ["PEGGY_TRACE"]` to the parser build task.

Verified both directions rather than taking the report on faith: the task hash was byte-identical with and without the variable (`34fe01c2095872cd` either way) and now differs (`f90d5b40...` vs `71c6bfa8...`), and the two generated grammars really are different files (`c14f0ca3...` untraced vs `1badfc3f...` traced).

Worth noting this hazard was introduced by this PR: while `src/grammar.js` was regenerated by executing the task, a missing env input was harmless. Declaring it a restorable output is what made a traced and an untraced build interchangeable in the cache.

Issue #758, opened 2026-06-08
Whitespace-only blank line in .env crashes parser (`expected dataType to be set`)


yinjs:
## Summary

A `.env` file containing a line with only spaces or tabs dead-ends the parser. On `varlock@1.4.0` it surfaces as a raw crash:

```
Error: expected dataType to be set
```

On newer versions the same input surfaces as a parse error. Standard dotenv parsers tolerate whitespace-only blank lines, so varlock should too.

## Reproduction

`.env`:
```
FOO=foo
   
BAR=bar
```
(middle line contains only spaces/tabs)

```
varlock load
```

→ parse fails / `expected dataType to be set`.

## Affected cases

- Whitespace-only line mid-file
- Trailing whitespace-only line at EOF with no final newline
- Multiple consecutive whitespace-only lines
- Whitespace-only line directly after a comment or decorator line

## Root cause

The PEG grammar (`packages/env-spec-parser/grammar.peggy`) blank-line rule matched only a bare `"\n"`, and the line terminator `_n` did not tolerate trailing horizontal whitespace — so a line of spaces/tabs had no valid production.

## Fix

PR #757 — three grammar changes:
- `_ "\n"` — whitespace-then-newline blank line
- `__ !.` — whitespace-only final line at EOF, no trailing newline
- `_n = _ ("\n" / !.)` — terminator tolerates trailing whitespace (fixes post-comment/decorator case)

Plus 7 regression tests. `@env-spec/parser` 217 pass, `varlock` 507 pass.

## Environment

- varlock 1.4.0 (originally reported); reproduced on `main`

Issue #720, opened 2026-05-26
[BUG]: @initGsm(projectId=$VAR) — dynamic variable reference silently fails, surfaces as misleading "expected dataType to be set"


coltmans:
## Describe the bug

When passing an environment variable reference as the `projectId` argument to `@initGsm`, varlock silently fails to initialize the GSM plugin instance. Rather than surfacing a clear "plugin not configured" error, it later throws:

```
Error: expected dataType to be set
```

from `config-item.ts:567`, which gives no hint that the root cause is the `@initGsm` argument.

## Steps to reproduce

```
# @plugin(@varlock/google-secret-manager-plugin)
# @initGsm(projectId=$GOOGLE_CLOUD_PROJECT)

MY_SECRET=gsm("my-secret-name")
```

Run `npm run dev` (or any `next build`) with `GOOGLE_CLOUD_PROJECT` set in the environment.

**Expected**: GSM plugin initializes with the resolved project ID and fetches the secret.
**Actual**: `Error: expected dataType to be set`

## What we tried

We wanted to make the project ID dynamic so the same schema works across multiple GCP environments. We consulted:

- Plugin README: https://github.com/dmno-dev/varlock/tree/main/packages/plugins/google-secret-manager
- Plugin docs (Setup + Auth section): https://varlock.dev/plugins/google-secret-manager/
- Multiple plugin instances docs: https://varlock.dev/plugins/google-secret-manager/#multiple-plugin-instances

And tried every approach we could find:

- `@initGsm(projectId=$GOOGLE_CLOUD_PROJECT)` — fails with `expected dataType to be set`
- `@initGsm(projectId=$MY_PROJECT_VAR || "my-fallback-project")` — fails with `expected dataType to be set`
- `@initGsm(projectId=$MY_PROJECT_VAR)` with `MY_PROJECT_VAR` defined as `@public @type=string` earlier in the same schema — fails with `expected dataType to be set`

The [README API reference](https://github.com/dmno-dev/varlock/tree/main/packages/plugins/google-secret-manager#initgsm) shows that `credentials` supports `$VAR` syntax (e.g. `credentials=$GCP_SA_KEY`), so the expectation is that `projectId` should work the same way.

## Workaround

Use a named instance with a static project ID and pass the instance ID explicitly to every `gsm()` call (per the [multiple instances docs](https://varlock.dev/plugins/google-secret-manager/#multiple-plugin-instances)):

```
[...]

Issue #629, opened 2026-04-18
[BUG]: Plugin load failure surfaces as expected dataType to be set assertion instead of the actual error


schalkneethling:
### varlock version

0.9.0

## Summary

When `@varlock/1password-plugin` fails to load (in my case due to a version mismatch with `varlock` core), schema resolution continues and eventually trips an internal assertion at `ConfigItem.resolve` with `Error: expected dataType to be set`. The original plugin load error is never surfaced, which makes the real cause very hard to find.

Separately, the two published plugin packages I use declare peer ranges that don't include `varlock@0.8.x` or `0.9.x`, so a fresh install of `varlock@latest` alongside the plugins produces a broken combination by default.

## Environment

- `varlock@0.9.0`
- `@varlock/1password-plugin@0.2.1` (initially) / `0.3.5` (after update)
- `@varlock/vite-integration@0.2.10`
- Node.js 24.15.0
- pnpm 10.33.0
- Reproduced both locally (macOS) and on Netlify CI

## Reproduction

`.env.schema`:

```
# @defaultRequired=infer @defaultSensitive=false
# @generateTypes(lang=ts, path=env.d.ts)
# @plugin(@varlock/1password-plugin)
# @initOp(allowAppAuth=true, account=A1B2C3D4E5F6G7H8)
# ---

# @type=url(startsWith=libsql://) @required
TURSO_DATABASE_URL=op(op://dev/my-db/db-url)

# @sensitive @required
TURSO_AUTH_TOKEN=op(op://dev/my-db/auth-token)
```

With `varlock@0.9.0` + `@varlock/1password-plugin@0.2.1`, run `npx varlock load`.

## Expected behaviour

When plugin loading fails, varlock should surface that error at the point of failure and halt, rather than continuing into resolution where an internal invariant trips with no context.

The plugin `loadPluginModuleCJS` path attempting `new Function(...)` on module source that contains ESM `import` statements suggests the loader is choosing the wrong module format branch for the plugin's published artifact. Whatever the right fix is there (detection, both-format support, clearer error), the downstream assertion at `ConfigItem.resolve` should not be the user's first signal.

## Root cause (for the case I hit)

Peer-range drift between core and plugins:

- `varlock@0.9.0` is the latest core
- `@varlock/1password-plugin@0.3.5` (latest) declares `peerDependencies: { varlock: "^0.7.3" }`
[...]

PR #1069, opened 2026-09-05
fix(varlock): record missing dataType as a ResolutionError

hassaans:
## Summary

`ConfigItem.resolve` throws `Error: expected dataType to be set` when `finishLoad()` returns early (any invalid source, or a plugin `loadingError`) and the item is then resolved because its key is also in `process.env` / `overrideValues`.

`EnvGraph.resolveEnvValues()` does not await `resolveItem()` and uses a dead `_reject`, so that throw becomes an **unhandledRejection** while the promise the caller awaits hangs forever.

This is the crash hq-cli users hit as Sentry `7687037996` (and the same surface as #629 / #720 / #758). No released version removes the assertion.

## Change

Record a `ResolutionError` on the item and return, matching every other resolve failure. `resolveEnvValues()` can then settle; callers see `item.errors` instead of a floated assertion.

## Test

`packages/varlock/src/env-graph/test/unprocessed-item-resolve.test.ts` — valid `.env.schema` + malformed `.env.local` + override. Without the fix, `resolveEnvValues()` hangs and the assertion is unhandled. With it, the promise settles and the item carries a `ResolutionError`.

Does not change successful resolution.

Related: https://github.com/indigoai-us/hq-cli (consumer workaround while this lands).

### DIFF
--- a/packages/varlock/src/env-graph/lib/config-item.ts
+++ b/packages/varlock/src/env-graph/lib/config-item.ts
@@ -1076,7 +1076,19 @@ export class ConfigItem {
     }
 
     const dataType = this.effectiveDataType;
-    if (!dataType) throw new Error('expected dataType to be set');
+    // finishLoad() returns early when any source is invalid or a plugin failed
+    // to load, leaving items unprocessed (dataType unset) and with no schema
+    // error. Resolving such an item — typically because the key is also in
+    // process.env, which varlock treats as an override — used to throw
+    // `expected dataType to be set` from an un-awaited resolveItem(), which
+    // becomes an unhandledRejection while resolveEnvValues() hangs (dead
+    // `_reject`). Record it as a ResolutionError so callers can fail cleanly.
+    if (!dataType) {
+      this.resolutionError = new ResolutionError(
+        'expected dataType to be set — this item was never typed because env-graph loading did not finish. Check .env / .env.schema / .env.local for parse errors, and that plugins loaded.',
+      );
+      return;
+    }
 
     // COERCE VALUE - often will do nothing, but gives us a chance to convert strings to numbers, etc
     try {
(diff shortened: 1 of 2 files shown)


theoephraim:
Thanks for this. Will get it sorted asap.

theoephraim:
Thanks for the report and the repro, that was a good catch and the diagnosis is exactly right.

We went a different direction on the fix though, in #1072.

`expected dataType to be set` is an internal invariant guard rather than a user-facing error. It's only reachable because we resolved a graph we already knew had failed to load: `finishLoad()` bails before typing any items when a source fails to parse or a plugin fails to install. Turning it into a `ResolutionError` would make the crash survivable but leave the user staring at a message about an item's type, when the real problem is a parse error in a file we could point at with a line number.

So #1072 catches it a level earlier instead:

- `checkForSchemaErrors()` now runs before resolution in the places that were missing it (the exported `load()`, plus `encrypt --file`, `keychain import`, and `scan`). Most CLI commands already did this, which is why the crash never showed up through `varlock load` or `varlock run`. You get the parse error with file and line instead of a crash.
- `resolveEnvValues()` refuses to run at all on a graph whose load bailed, so a call site that forgets fails immediately with something actionable.
- The other half of your diagnosis, the dead `_reject`, is fixed. Item resolution is fire-and-forget, so any throw became an unhandledRejection *and* left the returned promise permanently pending. That was a real bug on its own and is the reason this presented as a hang rather than an error.
- The assertion itself stays a plain `Error`.

Two questions, if you don't mind:

**How are you using varlock in hq-cli?** I'd like to make sure the fix actually covers your path. The crash reaches `ConfigItem.resolve()` the same way regardless, but knowing whether you're calling the exported `load()`, going through `internal.loadVarlockEnvGraph()` yourself, or something else tells us where to put the guard rails. #1072 also adds `checkForSchemaErrors` to the `internal` export, since that surface previously handed you `loadVarlockEnvGraph` and `checkForConfigErrors` but not the check that has to run between them.

**Heads up that `load()` may be deprecated.** It isn't documented anywhere and predates the current integration story. If that's what you're on, `varlock run` or `import 'varlock/auto-load'` are the supported paths, and we'd want to make sure whatever you need from `load()` is covered before removing it.

Closing this in favor of #1072, but the fix is yours in substance. Thanks again.

PR #1079, opened 2026-09-09
fix(varlock): stop probing the native encryption helper on every load

### LINKED ISSUE #1078: WSL: the win32 native helper is invoked on every run and costs ~5 s, even when nothing is encrypted
Running varlock inside WSL 2 costs about 5.9 seconds per invocation, because the binary
resolver maps WSL to the Windows helper and then executes `varlock-local-encrypt.exe`
through WSL interop. Our `.env.schema` contains no encrypted values at all, so the helper
is doing work we never asked for, and it dominates the runtime of every command.

## Environment

- varlock 1.18.0, `@varlock/native-helper-win32-x64@1.18.0` and
  `@varlock/native-helper-linux-x64@1.18.0` both installed
- Bun 1.4.0, Ubuntu 22.04 on WSL 2, Windows 10, repository on the ext4 side (`/home/...`)

## What happens

`VARLOCK_DEBUG=1` shows the resolution and the interop call:

```
[varlock:binary-resolver] resolving: platform=linux, isWSL=true, binaryName=varlock-local-encrypt.exe, subdir=win32-x64
[varlock:binary-resolver] resolved via platform package: .../@varlock/native-helper-win32-x64/varlock-local-encrypt.exe
[varlock:local-encrypt] detectBackendType: binaryPath=.../varlock-local-encrypt.exe, isWSL=true, platform=linux
[varlock:local-encrypt] runNativeBinary: .../varlock-local-encrypt.exe status
```

The `status` probe is the expensive part: launching a Windows executable that lives on the
Linux filesystem through interop is slow, and it happens on every single run.

## Measurements

Command, repeated three times per configuration:

```
bun --no-env-file node_modules/varlock/bin/cli.js run --inject vars --path .env.schema -- sh -c 'exit 0'
```

| Configuration | Time |
| --- | --- |
| Default install | 5758 / 5900 / 5922 ms |
| `_VARLOCK_FORCE_FILE_ENCRYPTION_FALLBACK=1` | 622 / 889 / 908 ms |
| `@varlock/native-helper-win32-x64` deleted from `node_modules` | 829 / 829 / 1011 ms |

The practical effect on our repository: a test file that shells out to varlock nine times
takes 34 s instead of 6.5 s, and several tests exceed Bun's default 5000 ms timeout, so the
suite is red on WSL and green everywhere else.

## What we expected

The Windows helper only exists to protect locally encrypted values with DPAPI and Windows
Hello. When the schema declares no encrypted values, we expected varlock never to touch it.

Possible directions, in the order we would find useful:

1. Skip the `detectBackendType` / `status` probe entirely when the loaded schema contains no
   encrypted values and no operation needs the local-encrypt backend.
2. Under WSL, prefer `@varlock/native-helper-linux-x64` when it is installed, and reach for
   the Windows helper only when DPAPI or Windows Hello is actually required.
3. Cache the backend status per project instead of probing on each invocation.
4. Failing all of the above, promote `_VARLOCK_FORCE_FILE_ENCRYPTION_FALLBACK` to a
   documented public option, so projects can opt out without relying on an internal name.

## Workaround we are using

Exporting `_VARLOCK_FORCE_FILE_ENCRYPTION_FALLBACK=1` when the shell detects WSL. It works,
but it is an underscore-prefixed internal variable, so we would rather not depend on it.

philmillman:
Fixes #1078

## What

On WSL 2, every `varlock load` / `varlock run` took ~5.9s instead of ~0.7s, even for schemas with no encrypted values. Debug output showed `varlock-local-encrypt.exe status` being spawned on each invocation.

The eager spawn was not the `varlock()` resolver. The cache auto-policy in the graph loader called `getBackendInfo()` before parsing any .env file, only to check whether a native backend exists. That function runs the helper's `status` command, which on WSL is a Windows .exe launched through interop.

## Changes

- Add `getBackendType()`: a filesystem-only classification (is a native helper installed, which platform) that never spawns the helper.
- The loader's cache policy and the `@cache=disk` root decorator branch use `getBackendType()`. The loader also checks the CI flag first, so CI skips even the filesystem check.
- `getBackendInfo()` (the full `status` probe) is unchanged, but now only runs lazily from operations that need its output: key management, encrypt/decrypt, `varlock(prompt)`, `varlock encrypt`, `varlock lock`. The WSL `cachedStatusKeys` shortcut for `keyExists` still works because those paths call `getBackendInfo()` first.
- Regression test: the loader cache-policy test now fails if `getBackendInfo()` is ever called during a load.
- One sentence in the WSL section of the local encryption guide.

Not changed: WSL still maps to the Windows helper (intentional, for DPAPI/TPM + Windows Hello), and the `status` result is not persisted across invocations. Projects that do decrypt on WSL still pay the probe once per run; that could be a separate follow-up.

## Verification

- Full varlock vitest suite, `tsc --noEmit`, and lint pass.
- Manual check on macOS with `VARLOCK_DEBUG=1`: a plain schema logs only the filesystem check and no `status` spawn; a schema with a `varlock("local:...")` value still triggers `status` exactly once right before decrypt.
- Not tested on WSL itself. @tigerxml, if you can re-run your timing on this branch that would confirm the fix end to end.


philmillman:
I'll open a separate PR for the `audit` issues, which will probably require an astro 7 upgrade @theoephraim

PR #1092, opened 2026-09-16
fix(polkit): correct Linux biometric setup command in hint

denniseijpe:
This is just a simple fix for the setup hint polkit prints.

`setup-linux-biometrics` doesn't exist on `varlock-local-encrypt`.

`setup --linux-biometrics` does.


theoephraim:
Thanks! Will have this out in next release.

